PR #4, opened 2018-10-12
Version 1.0 of the template with working validation test

n01d:
Version 1.0 of the template with working validation test, full test still pending.


dcaro:
The module looks god but we need to add outputs to make this module more usable.

PR #18, opened 2019-05-25
added kubeconfig output

jungopro:
This PR adds an option to output the kubeconfig to a file on the local file system
A user using the module will then need to add the file to his local KUBECONFIG environment variable (as explained in the terraform variable description) and will be able to connect to the cluster

This saves the need to further run `az aks get-credentials ...` post module execution


n01d:
Hallo @jungopro , thanks for your input and sorry it took me so long to come back to this. Could you please update to terraform 0.12 syntax and resolve the conflicts so I can test and merge your addition to the module?

PR #35, opened 2020-01-20
Update the template validation pipeline to use the latest testing image

n01d:
Update the template validation pipeline to use the latest testing image. The new image image is pulled from the Microsoft container registry and contains 0.12 which is required by the template.


yupwei68:
Hi @n01d ,Thanks for opening this PR. Sorry that I have include that in the latest release, and I have to close the PR.

PR #39, opened 2020-02-05
terraform version upgrade and code reorg

yupwei68:


README.md
```diff
@@ -6,87 +6,103 @@ This Terraform module deploys a Kubernetes cluster on Azure using AKS (Azure Kub
...
-}
+```sh
+$ curl -sSL https://raw.githubusercontent.com/Azure/terramodtest/master/tool/env_setup.sh | sudo bash
 ```
 
-The module supports some outputs that may be used to configure a kubernetes
-provider after deploying an AKS cluster.
+#### Run test
 
+Then simply run it in local shell:
```
mybayern1974: Please add more instructions for run test locally

Issue #41, opened 2020-02-17
update agent_pool_profile for default_node_pool


framled:
A agent_pool_profile block will be removed in 2.0, this should be changed for default_node_pool

yupwei68:
hi @framled ,Thanks for your remind. We'll fix that in the next release.

Issue #29, opened 2019-11-08
Preview features


ekjuanrejon:
Hello

When will this support the node pool?

Can we disable log analytics

sblack4:
I've got a fork that uses node pool. If you submit a PR to this or that repo which adds the ability to disable log analytics I'll approve it 
https://github.com/rhythmictech/terraform-azurerm-aks

Issue #34, opened 2020-01-20
Terraform plan output is the recreation of AKS Cluster


nidhi5885:
I have implemented this modular approach in creation of AKS. Terraform Plan and Terraform Apply did their work, and I got my cluster up and running. But then by not doing any change to my AKS cluster, I executed Terraform plan again and as per it's output my AKS Cluster need to recreated(I don't know why it is happening, or why the Terraform is unable to match the states) 
To reproduce the issue: You just need to fork the repo,  make some required changes, execute terraform plan and terraform apply, when you see that the cluster is up and running. Again run terraform plan(without changing anything to the cluster)

nidhi5885:
Any update on this? Help required.

MichaelMcCoyTech:
I see this problem too!

jmhill:
I ran into this problem as well (ran a terraform config that was calling only this module to test it out, and then after applying the terraform config immediately ran a `terraform plan` to verify everything was good), and noted that the cluster block in the plan had several fields marked as forcing the cluster to be recreated.

I don't have time to mess with it too much right now, but looking at the plan it seemed to me in my case like there was a problem with some sort of mapping between the deprecated `agent_pool_profile` argument that is used by this module, and the `default_node_pool` that is supposed to be used in newer versions of the azurerm provider (see https://www.terraform.io/docs/providers/azurerm/r/kubernetes_cluster.html#agent_pool_profile). At least in my case, the `terraform plan` output showed an attempt to change those fields to null on subsequent runs, when I had changed nothing about the config.

In my case I got around this by just using my own config from the tutorial (https://docs.microsoft.com/en-us/azure/terraform/terraform-create-k8s-cluster-with-tf-and-aks#define-a-kubernetes-cluster) which uses the new `default_node_pool`, and I had no problems. I suspect if you're using a newer version of terraform/azurerm provider you could fork this repo and change `agent_pool_profile` in the main.tf to `default_node_pool` (and update any of the arguments within that block as needed) and it'd probably work.

yupwei68:
HI @nidhi5885 ,@jmhill , Thanks for opening this issue. We'll use default_node_pool in the next release v4.0 to support Azurerm 2.0. This problem will not appear in the release v4.0.

PR #42, opened 2020-03-13
Support Azurerm 2.0

yupwei68:
1. support Azurerm 2.0
fix:
https://github.com/Azure/terraform-azurerm-aks/issues/41
https://github.com/Azure/terraform-azurerm-aks/issues/34

2. Allow to disable log analytics
fix https://github.com/Azure/terraform-azurerm-aks/issues/29


main.tf
```diff
@@ -34,10 +34,13 @@ resource "azurerm_kubernetes_cluster" "main" {
     client_secret = var.client_secret
   }
 
-  addon_profile {
-    oms_agent {
-      enabled                    = true
-      log_analytics_workspace_id = azurerm_log_analytics_workspace.main.id
+  dynamic addon_profile {
+    for_each = var.enable_log_analytics_workspace ? ["log_analytics"] : []
+    content {
```
mybayern1974: I may miss some tech background but do we need this "content{}" block? Can we just have the "ome_agent{}" within the "addon_profile{}" w/o the "content{}"?

main.tf
```diff
@@ -46,19 +49,23 @@ resource "azurerm_kubernetes_cluster" "main" {
...
 }
 
 resource "azurerm_log_analytics_solution" "main" {
+  count                 = var.enable_log_analytics_workspace ? 1 : 0
   solution_name         = "ContainerInsights"
   location              = data.azurerm_resource_group.main.location
   resource_group_name   = var.resource_group_name
-  workspace_resource_id = azurerm_log_analytics_workspace.main.id
-  workspace_name        = azurerm_log_analytics_workspace.main.name
+  workspace_resource_id = azurerm_log_analytics_workspace.main[0].id
```
mybayern1974: What if the count is zero? Will there be any friendly error message / default behaviors, or anything will crash?

main.tf
```diff
@@ -46,19 +49,23 @@ resource "azurerm_kubernetes_cluster" "main" {
...
+  workspace_name        = azurerm_log_analytics_workspace.main[0].name
```
mybayern1974: The same as the above

test.sh
```diff
@@ -1,35 +0,0 @@
-#! /bin/bash
```
mybayern1974: The reason of removing this file is? Or to ask do we possibly know the original purpose of adding this file?

variables.tf
```diff
@@ -45,8 +45,13 @@ variable "public_ssh_key" {
...
-  default     = {}
-  description = "Any tags that should be present on the Virtual Network resources"
   type        = map(string)
+  description = "Any tags that should be present on the Virtual Network resources"
+  default     = {}
 }
 
+variable "enable_log_analytics_workspace" {
+  type        = bool
+  description = "Enable creation of azurerm_log_analytics_workspace and azurerm_log_analytics_solution?"
```
mybayern1974: Assume this is a typo issue of using a question mark "?"

main.tf
```diff
@@ -34,10 +34,13 @@ resource "azurerm_kubernetes_cluster" "main" {
...
-  addon_profile {
-    oms_agent {
-      enabled                    = true
-      log_analytics_workspace_id = azurerm_log_analytics_workspace.main.id
+  dynamic addon_profile {
+    for_each = var.enable_log_analytics_workspace ? ["log_analytics"] : []
+    content {
+      oms_agent {
+        enabled                    = true
+        log_analytics_workspace_id = azurerm_log_analytics_workspace.main[0].id
```
mybayern1974: What if the count is zero? Will there be any friendly error message / default behaviors, or anything will crash?

main.tf
yupwei68: No. Content is required in dynamic blocks.

main.tf
yupwei68: It will not be zero, it is created in the same condition with log analytics workspace. Two cases have been test in fixture test, enable log analytics or not.

main.tf
yupwei68: The same as above, too

test.sh
yupwei68: It's used to test the fixture and integrate CI. But we have used our ruby framework to integrate CI. I have left it not used at last release. But it seems it causes confusion for someone read the code. So I decided to delete it.

main.tf
yupwei68: In this case, it will not be zero. I could add something conditional equation. But the field can not be assigned with empty string anyway. I have test them in two conditions in fixture.

PR #46, opened 2020-05-10
Added a output for the raw kube_config to enable easier use with kubectl

lukekhamilton:


dargor:
This would be quite useful, and even expected given the [official documentation](https://docs.microsoft.com/en-gb/azure/developer/terraform/create-k8s-cluster-with-tf-and-aks#test-the-kubernetes-cluster) use a similar mechanism.

PR #55, opened 2020-10-02
Add support for AKS paid SKU

ThorstenHans:
This PR adds support for AKS paid SKU. 

The default value of `sku_tier` is set to `Free` to ensure backward compatibility

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ module "aks" {
   prefix              = "prefix"
   vnet_subnet_id      = module.network.vnet_subnets[0]
   os_disk_size_gb     = 50
+  enable_azure_policy = true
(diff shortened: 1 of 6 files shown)


yupwei68:
Hi @ThorstenHans , thanks for opening this PR. Overall it looks fine to me. Could you please 1. add this feature into fixture/test, and test if works when it's assigned "Paid"  2. add this feature into the example of REAME.

ThorstenHans:
@yupwei68 

I have added an AKS instance to the test fixture and verified creation by checking the resulting AKS id (ARM doesn't expose SKU tier on AKS resource).

Additionally, I have added the `sku_tier` to the sample shown in the README as suggested.

yupwei68 (approved):
LGTM!

PR #56, opened 2020-10-28
add variable to enable_kube_dashboard

### LINKED ISSUE #47: kube_dashboard is enabled, but provider can't disable it.
# Problem
`terraform apply` to spawn a brand new k8s cluster in a brand new resource group works fine, however each successive `terraform apply` wants to make the following modication to the cluster, even if I change nothing in my HCL files :
```
[...]
      ~ addon_profile {
          - kube_dashboard {
              - enabled = true -> null
            }
[...]
```
This does not seem to be manageable using module variables, or I missed it.

# Criticity
Mostly an annoyance as everything else works fine, would be fine to fix though.

# Versions
```
Terraform v0.12.26
+ provider.azurerm v2.15.0
+ provider.local v1.4.0
+ provider.tls v2.1.1
```

ANGkeith:
closes #47 

add variable to enable/disable the kube_dashboard.

defaults to `false`

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -92,6 +92,12 @@ variable "enable_azure_policy" {
   default     = false
 }
 
+variable "enable_kube_dashboard" {
+  description = "Enable Kubernetes web dashboard. NOTE: This add-on is set for deprecation. [https://docs.microsoft.com/en-us/azure/aks/kubernetes-dashboard]"
+  type        = bool
+  default     = false
+}
+
 variable "sku_tier" {
   description = "The SKU Tier that should be used for this Kubernetes Cluster. Possible values are Free and Paid"
   type        = string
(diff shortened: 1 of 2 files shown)


variables.tf
```diff
@@ -92,6 +92,12 @@ variable "enable_azure_policy" {
   default     = false
 }
 
+variable "enable_kube_dashboard" {
+  description = "Enable Kubernetes web dashboard. NOTE: This add-on is set for deprecation. [https://docs.microsoft.com/en-us/azure/aks/kubernetes-dashboard]"
+  type        = bool
+  default     = false
```
yupwei68, line 98: default = true

yupwei68 (commented):
Thanks for contribution. Except this minor change, would you mind add this feature into the test in test/fixture and in the example in README?

yupwei68:
Sorry that I've to close this PR, for https://github.com/Azure/terraform-azurerm-aks/pull/74 has been merged.

PR #61, opened 2020-10-31
Add default _node_pool dynamic config block 

bamaralf:


bamaralf:
@mattetti @jdhuntington @nmeans @bratsche

bamaralf:
Hi @yupwei68. I kept the module compatible with the previous variables. You will be able to use the "default_node_pool" var or the previous ones.

PR #67, opened 2020-11-08
Add support for private cluster API access.

es1o:
Hello, this is my first PR for this project. I've started using this module recently to manage my AKS cluster in my infrastructure.

I've added new variable because I'd like to restrict cluster access from internet in my infrastructure.

This change doesn't broke module behaviour because default is `false`. Default AKS cluster installation exposes API over internet as before.

Please let me know how do you think about this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,8 +45,7 @@ module "aks" {
   enable_role_based_access_control = true
   rbac_aad_admin_group_object_ids  = [data.azuread_group.aks_cluster_admins.id]
   rbac_aad_managed                 = true
-
-  private_cluster_enabled = false # default value
+  private_cluster_enabled          = true # default value
 
   depends_on = [module.network]
 }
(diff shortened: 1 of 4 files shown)


yupwei68:
Hi @es1o  Thanks for contribution. Would you mind adding this feature the test in test/fixture and in the example in README?

es1o:
@yupwei68 Ok, I'll add. :)

es1o:
@yupwei68 I added new variable to example.

I decided to not write tests because it requires to spawn new cluster. I didn't add any new logic, only pass variable to terraform aks resource as is. What do you think?

yupwei68:
@es1o , you don't have to add new cluster. You can just add one field in test module "aks". Thus, I can integrate this enhancement into CI to test if it works with this field in module "aks",  and if it works without this field ("aks-without-monitor"), and to avoid being broken by following changes.

es1o:
@yupwei68 Done, I've added to tests.

yupwei68 (approved):
LGTM!

PR #71, opened 2020-11-30
Set minimal provider required version

mcanevet:
Support for `tags` in `azurerm_log_analytics_solution` has been added in v2.34.0


yupwei68 (approved):
Thanks! LGTM!

Issue #66, opened 2020-11-06
Proposal - add cluster autoscaler support


es1o:
Hello.
I'd like to use AKS cluster autoscaler to scale nodes in my cluster.

Could I create Pull Request with autoscaler support? Terraform module supports cluster autoscaling.

What do you think?

yupwei68:
Hi @es1o , thanks for opening this issue. You're welcome to contribute to the module. Please add features into test/fixture to integrate CI and add features in example of README.

PR #75, opened 2020-12-01
Add support for nodepool autoscaling

lays147:

Add support for nodepool autoscaling with the following configuration to the module:
```hcl
  enable_auto_scaling  = true
  agents_min_count     = 1
  agents_max_count     = 2
```
Note: To build the custom image to run the tests I needed to comment the following lines on the Dockerfile:
```dockerfile
# RUN mkdir /go
# RUN mkdir /go/bin
# RUN mkdir /go/src
```
Otherwise, the following error would happen:
```
 ---> Using cache
 ---> 46c203d86fc0
Step 18/28 : RUN mkdir /go
 ---> Running in e91071d40948
mkdir: cannot create directory ‘/go’: File exists
The command '/bin/sh -c mkdir /go' returned a non-zero code: 1
```
The base image already has the `go` folders in it.

### DIFF
--- a/test/fixture/main.tf
+++ b/test/fixture/main.tf
@@ -43,7 +43,8 @@ module "aks" {
   enable_auto_scaling             = true
   agents_min_count                = 1
   agents_max_count                = 2
-  depends_on                      = [azurerm_resource_group.main]
+
+  depends_on = [azurerm_resource_group.main]
 }
   
 module "aks_without_monitor" {
--- a/test/fixture/main.tf
+++ b/test/fixture/main.tf
@@ -43,6 +43,7 @@ module "aks" {
   enable_auto_scaling             = true
   agents_min_count                = 1
   agents_max_count                = 2
+  agents_count                    = null
 
   depends_on = [azurerm_resource_group.main]
 }
--- a/test/fixture/main.tf
+++ b/test/fixture/main.tf
@@ -46,7 +46,7 @@ module "aks" {
 
   depends_on = [azurerm_resource_group.main]
 }
-  
+
 module "aks_without_monitor" {
   source                         = "../.."
   prefix                         = "prefix2-${random_id.prefix.hex}"
--- a/test/fixture/main.tf
+++ b/test/fixture/main.tf
@@ -38,6 +38,17 @@ module aks {
   depends_on                      = [azurerm_resource_group.main]
 }
 
+module aks_with_scaling {
+  source              = "../.."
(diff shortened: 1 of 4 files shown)


lays147:
Hi @yupwei68, is there anything else that I need to do for this PR to be approved?

test/fixture/main.tf
```diff
@@ -38,6 +38,17 @@ module aks {
   depends_on                      = [azurerm_resource_group.main]
```
yupwei68, line 38: Here, we add 
```
  enable_auto_scaling = true
  agents_min_count    = 1
  agents_max_count    = 2
```

variables.tf
```diff
@@ -47,7 +47,7 @@ variable "log_retention_in_days" {
 variable "agents_count" {
   description = "The number of Agents that should exist in the Agent Pool"
```
yupwei68, line 48: We want to add a description like: Please set `agents_count` `null` while `enable_auto_scaling` is `true` to avoid possible `agents_count` changes.

yupwei68 (commented):
Thanks for your contribution. 🚀  This PR looks almost good to me. I had intention to add a dynamic ignore_changes for "agents_count", but it's not supported from HCL. Just minor changes, it will be good for me. Thanks

yupwei68:
Fix https://github.com/azure/terraform-azurerm-aks/issues/66

test/fixture/main.tf
```diff
@@ -38,6 +38,17 @@ module aks {
...
 }
 
+module aks_with_scaling {
+  source              = "../.."
+  prefix              = "prefix-${random_id.prefix.hex}"
+  resource_group_name = azurerm_resource_group.main.name
+  enable_auto_scaling = true
+  agents_min_count    = 1
+  agents_max_count    = 2
```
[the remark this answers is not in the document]
lays147: Yes, working on it.

variables.tf
lays147: Well, we are not able to ignore change to a variable, but we can ignore the change to the resource variable that are receiving that value:
```hcl
  lifecycle {
    ignore_changes = [default_node_pool["node_count"]]
  }```

What do you think?

yupwei68: It'll block users when they want to update the "node_count" while auto_scaling disabled.

yupwei68 (approved):
LGTM! On second consideration, I don't change the default value of `agents_count` from 2 into null, since it'll break our users. Thanks!

PR #79, opened 2020-12-14
Default node pool

bamaralf:
Fixes #000 

- Remove the "-aks" from the prefix var
- Add the available configs to the 'default_node_pool' config block


omerfsen:
Wow 7 months of commits ;)

PR #82, opened 2020-12-22
Change Version Constraint to allow for minor Azurerm Provider releases

### LINKED ISSUE #81: Changing the Azurerm Provider Version Constraint
Is it possible to change the required version constraint from `~> 2.34.0` to `~> 2.34`? At the moment, having `2.34.0` is basically the same as being fixed as only the rightmost version component is incremented and the azurerm provider always tends to be released in minor increments. 

If you try and use a higher provider elsewhere e.g. 2.41.0, you'll see an error similar to:

```
Could not retrieve the list of available versions for provider
hashicorp/azurerm: no available releases match the given constraints ~>
2.34.0, 2.34.1
```

matthorgan:
Fixes #81 

Removes patch version constraint so that minor versions of new azurerm providers are allowed.

### DIFF
--- a/versions.tf
+++ b/versions.tf
@@ -3,7 +3,7 @@ terraform {
   required_providers {
     azurerm = {
       source  = "hashicorp/azurerm"
-      version = "~> 2.34.0"
+      version = "~> 2.34"
     }
   }
 


yupwei68 (approved):
LGTM! Thanks!

PR #89, opened 2021-01-27
added suffix, fixed names to match msft recomendations, fixed names o…

AlexBevan:
This is a big breaking change due to changing the name of the terraform resources to comply with terraform best practices.

- Renamed resources to match MSFT naming guidelines
- Added support for suffix
- Added support to add node pools

Fixes #000 


raphink:
@yupwei68 is this change breaking solely n'existe l because of the `main` vs `this` renaming? If so, is it more likely to be merged by keeping the old name?

PR #92, opened 2021-02-22
Do not set default node pool node_count when auto scaling is enabled

### LINKED ISSUE #91: The default node pool node_count is always reset with autoscaling
The default node pool node_count is set always, even if autoscaling is enabled.

maxgio92:
**Changes proposed in the pull request**

To make the `default_node_pool` nested block `dynamic`, with content based on value of `enable_auto_scaling` input variable. In example, the `node_count` parameter of `default_node_pool` should be set only when autoscaling is disabled.
Tests pass.

Fixes #91


yupwei68 (approved):
LGTM!

PR #101, opened 2021-05-26
Add support for default pool host-based encryption

jcorioland:
This PR adds support for default pool [host-based encryption](https://docs.microsoft.com/en-us/azure/aks/enable-host-encryption).

* there is a new `enable_host_encryption` variable, that default to `false` and can be turned to `true` to enable host-based encryption
* the documentation has been updated with this new flag
* the tests have been updated with this new flag
* the minimum AzureRM provider version has been moved to `2.46` which corresponds to the release of this property in the provider

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ module "aks" {
   enable_http_application_routing  = true
   enable_azure_policy              = true
   enable_auto_scaling              = true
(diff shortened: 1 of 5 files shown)


yupwei68 (approved):
LGTM!

PR #102, opened 2021-05-31
Replace data source lookup for "azurerm_resource_group" with variables to avoid resources recreation issue when re-applying the module

nitinnbisht:
FIxes #105 
Changes proposed in the pull request:
When testing in terraform v13,14,15 - The current module results in recreation of cluster in consecutive apply due to the unwanted diff generated by data source lookup of `location` from the `resource group`. The fix in the PR proposes using variable for location in-place of a data source lookup to resolve this issue.

The data source lookup responsible:
```
 data "azurerm_resource_group" "main" {  
  name = var.resource_group_name
  }
```
Terraform plan using the module results in certain resources to `force replacement`, PFB
```
# module.aks.azurerm_kubernetes_cluster.main must be replaced
-/+ resource "azurerm_kubernetes_cluster" "main" {
      - api_server_authorized_ip_ranges = [] -> null
      - enable_pod_security_policy      = false -> null
---
      ~ location                        = "westeurope" -> (known after apply) # forces replacement
---
    }

  # module.aks.azurerm_log_analytics_solution.main[0] must be replaced
-/+ resource "azurerm_log_analytics_solution" "main" {
---
      ~ location              = "westeurope" -> (known after apply) # forces replacement
---
        }
    }

  # module.aks.azurerm_log_analytics_workspace.main[0] must be replaced
-/+ resource "azurerm_log_analytics_workspace" "main" {
 ---
      ~ location                   = "westeurope" -> (known after apply) # forces replacement
 ---
    }
```

PR #106, opened 2021-06-08
Added support for specifying cluster location

### LINKED ISSUE #105: Cannot override cluster location
cluster location defaults to using resource group location, and there's no way to override this.
Creating PR to address this.

larryclaman:
Fixes #105 

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -10,7 +10,7 @@ module "ssh-key" {
 resource "azurerm_kubernetes_cluster" "main" {
   name                    = var.cluster_name == null ? "${var.prefix}-aks" : var.cluster_name
   kubernetes_version      = var.kubernetes_version
-  location                = data.azurerm_resource_group.main.location
+  location                = var.location ==null ? data.azurerm_resource_group.main.location : var.location
   resource_group_name     = data.azurerm_resource_group.main.name
   dns_prefix              = var.prefix
   sku_tier                = var.sku_tier
--- a/variables.tf
+++ b/variables.tf
@@ -289,3 +289,8 @@ variable "enable_host_encryption" {
   type        = bool
   default     = false
 }
+
+variable "location" {
+  description = "Location (Region) for AKS Cluster.  If not set, will use Resource Group location"
+  type        = string
+}


kdouglas-qarik (approved):
I'm seeing an issue with this where my plan/apply is destroying/recreating the cluster every time since the location is 'known after apply'. Would be great to get this PR merged to prevent this occuring.

larryclaman:
hmm, let me look into this.  I agree that 'as-is', the PR should not yet be merged.

nitinnbisht:
@larryclaman and @kdouglas-qarik:  I think this [PR](https://github.com/Azure/terraform-azurerm-aks/pull/102) already solves this and more, as the resource recreation issue due to data source lookup not only affects AKS creation but loganalytics resource as well. 

I understand the intention here is to allow creating AKS cluster in a different location, but the linked PR addresses that as well. Please have a look.

main.tf
```diff
@@ -10,7 +10,7 @@ module "ssh-key" {
 resource "azurerm_kubernetes_cluster" "main" {
   name                    = var.cluster_name == null ? "${var.prefix}-aks" : var.cluster_name
   kubernetes_version      = var.kubernetes_version
-  location                = data.azurerm_resource_group.main.location
+  location                = var.location ==null ? data.azurerm_resource_group.main.location : var.location
```
onlydole, line 13:
```suggestion
  location                = var.location == null ? data.azurerm_resource_group.main.location : var.location
```

onlydole: This update was working for me, but curious to see if you are noticing any create/delete cycles still, @larryclaman or @kdouglas-qarik!

kdouglas-qarik: I ended up dropping use of this module entirely and putting the required terraform resources directly in my own module instead, but based on what I saw previously, setting the location with an optional var should prevent the create/delete cycle yes.

lonegunmanb:
Sorry for the crazy late reply, thanks @larryclaman for opening this pr and thank you for your contribution. This feature has been implemented by #172 so I'm closing this pr now.

Issue #115, opened 2021-07-23
Support the `ingress_application_gateway` addon


wjam-hanzo:
There is currently no support for enabling the `ingress_application_gateway` addon on the AKS cluster. Please add support for enabling this addon.

jcorioland:
@davidkarlsen hi, I am an external contributor to this repo, not actively working on it on my side. But, the team who owns this confirmed to me the module is still active.

davidkarlsen:
> @davidkarlsen hi, I am an external contributor to this repo, not actively working on it on my side. But, the team who owns this confirmed to me the module is still active.

OK, good to know, on the other side, there seems to be no activity on the PRs or releases.

davidkarlsen:
I think this issue can be closed as the PR has been merged now.

lonegunmanb:
Hi @davidkarlsen , since we're still haven't a new release yet, do you think it's ok to close this issue now?

davidkarlsen:
I guess that depends on your policy - if you close issues when they are merged, or tagged? It has been merged.
Which brings us to release policy - what criteria brings those along?

PR #116, opened 2021-08-05
Add support for using application gateway for ingress

### LINKED ISSUE #114: Expose `os_disk_type` as a variable
I've just spun up a Kubernetes cluster in Azure using this module and am reviewing the advisor recommendations. One of the recommendations is to switch to [emphemeral OS disks](https://docs.microsoft.com/en-gb/azure/aks/cluster-configuration?WT.mc_id=Portal-Microsoft_Azure_Expert#ephemeral-os), which would be achieved by setting the `os_disk_type` argument for the `default_node_pool` block to `Ephemeral` from the default `Managed`.

Please add a variable to be able to switch OS disk types to make use of this feature.

wjam:
This adds support for using an application gateway for ingressing traffic into the cluster. The Application Gateway Ingress Controller, which will be deployed into the cluster, is granted the `Contributor` role to modify the gateway as it sees fit.

Note that the controller will need to be granted the `Microsoft.Network/publicIPAddresses/read` permission on the IP address associated with the gateway for any `ingress` Kubernetes resources to have their address correctly updated.

This also adds support for setting the O/S disk type to `Ephemeral`.

Fixes #114 #115

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -44,6 +44,7 @@ resource "azurerm_kubernetes_cluster" "main" {
       tags                   = merge(var.tags, var.agents_tags)
       max_pods               = var.agents_max_pods
       enable_host_encryption = var.enable_host_encryption
+      os_disk_type           = var.os_disk_type
     }
   }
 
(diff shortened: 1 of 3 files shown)


micgruber:
What is the status on supporting application gateway for ingress because currently we use the module and need this feature?

lebenitza:
Ping? We really need the disk type flag.

nlamirault:
Any news on this feature ?

zioproto:
This PR is not needed anymore because the changes were implemented in #139 

The issue #114 is closed and #115 should be closed already.

@lonegunmanb  Please close the PR without merging.

PR #127, opened 2021-10-19
Support for another node pools

nlamirault:
See: https://github.com/Azure/terraform-azurerm-aks/pull/89


nlamirault:
See a more recent PR : https://github.com/Azure/terraform-azurerm-aks/pull/127

PR #132, opened 2021-11-26
Open Service Mesh addon

nlamirault:
This PR add support for [Open Service Mesh for AKS](https://docs.microsoft.com/en-us/azure/aks/open-service-mesh-about)


main.tf
```diff
@@ -102,6 +102,10 @@ resource "azurerm_kubernetes_cluster" "main" {
       log_analytics_workspace_id = var.enable_log_analytics_workspace ? azurerm_log_analytics_workspace.main[0].id : null
     }
 
+    open_service_mesh {
+      enabled = var.enable_open_service_mesh
+    }
+
```
zioproto, line 108: Looking at the documentation this is not a block:
https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster#open_service_mesh_enabled

It should be a oneliner:

`open_service_mesh_enabled = var.enable_open_service_mesh`

lonegunmanb: @zioproto you're right. For now this module works with provider 2.x so this patch is correct, but we're close to merge a pr that upgrade provider to 3.x so I'm not sure whether we should continue work on 2.x. My thought is we upgrade to 3.x first then let's see what we can do for this pr.

main.tf
```diff
@@ -98,6 +98,8 @@ resource "azurerm_kubernetes_cluster" "main" {
     }
   }
 
+  open_service_mesh_enabled = var.enable_open_service_mesh
+
```
zioproto, line 102: @nlamirault you introduced a line with empty spaces at line 102. We are going to introduce very soon a CI that will lint the code, and this file will not pass the linting. Could you please remove the empty spaces ?

To avoid this pain in the future you can set your editor to automatically trim empty spaces at the end of lines when you save the files. Thank you

nlamirault: Sorry :-1:  I made the fix using Github UI ... I fix that asap.

lonegunmanb: Hi @zioproto agreed but I think we can accept this pr for this time since we'll submit a single pr to fix all the naming issues and style issues.

lonegunmanb (approved):
LGTM, thanks @nlamirault

lonegunmanb (changes requested):
Hi @nlamirault these changes LGTM but there're some conflicts because we've accepted some prs, would you please solve these conflicts so I can merge this pr? Thanks!

variables.tf
```diff
@@ -128,6 +128,12 @@ variable "enable_azure_policy" {
   default     = false
 }
 
+variable "enable_open_service_mesh" {
+  description = "Enable Open Service Mesh Addon."
```
lonegunmanb, line 132: Can we use the same description as [provider's document](https://github.com/hashicorp/terraform-provider-azurerm/blob/main/website/docs/r/kubernetes_cluster.html.markdown?plain=1#L152)?

variables.tf
```diff
@@ -128,6 +128,12 @@ variable "enable_azure_policy" {
...
+  type        = bool
+  default     = false
```
lonegunmanb, line 134: Would the default value `null` be better? A `false` default value will cause changes in plan when the users upgrade module's value.

lonegunmanb (changes requested):
Almost LGTM but two tiny issues, would you consider make change? Thanks @nlamirault !

lonegunmanb (approved):
Thanks @nlamirault , LGTM

PR #141, opened 2022-02-02
Update main.tf

hilbabt:
Add support in load_balancer_sku

in our lab we are working in privert Lab in Azure that dont have public ip,
i was need to change that molde like that : load_balancer_sku

to support it

Fixes #000 


jiaweitao001 (approved):
LGTM. Approved.

PR #144, opened 2022-02-21
Create private_ssh_key file with a optional flag

### LINKED ISSUE #134: Allow export of private ssh key to be optional.
Exporting of private ssh keys should be optional. Users who have AKS clusters integrated with AzureAD may not need or want private keys exported.

gungoren:
Fixes #134  

Create private_ssh_key file with a optional flag


gungoren:
This change was inspired by https://github.com/terraform-aws-modules/terraform-aws-eks/blob/v17.24.0/kubectl.tf#L2

PR #163, opened 2022-05-09
feat(taints): to add taint feature to aks module

### LINKED ISSUE #152: Allow taints for default (system) pool
Currently it's not possible to taint System pool (default agent's pool) within the module.

Please allow for the parameter: [only_critical_addons_enabled = bool] or at least just a taint block to be added, e.g. 

agent_taints = [
    "CriticalAddonsOnly=true:NoSchedule"
  ]

Existing nodepools cannot be tainted after creation, only the single nodes within them.

carvido1:
Fixes #152  

A feature to add taints to the default nodepool has been added. This will help to follow [Microsoft best practices for system nodepools](https://docs.microsoft.com/en-us/azure/aks/use-system-pools#system-and-user-node-pools) as we will be allowed to add the `CriticalAddonsOnly=true:NoSchedule` taint to the system nodepool.

### DIFF
--- a/CHANGLOG.md
+++ b/CHANGLOG.md
@@ -6,3 +6,15 @@ ENHANCEMENTS:
 * Include `node_resource_group` as variable ([#136](https://github.com/Azure/terraform-azurerm-aks/pull/136))
 
 BUG FIXES:
+
+## 4.16.0 (June 02, 2022)
+
+ENHANCEMENTS:
+
+* Added output for `addon_profile` ([#151](https://github.com/Azure/terraform-azurerm-aks/pull/151))
+* Adding Microsoft SECURITY.MD ([#167](https://github.com/Azure/terraform-azurerm-aks/pull/167))
+* Added variable `os_disk_type` for default node pools ([#169](https://github.com/Azure/terraform-azurerm-aks/pull/169))
+
+BUG FIXES:
+
+* Trivial fix to the example in the README ([#166](https://github.com/Azure/terraform-azurerm-aks/pull/166))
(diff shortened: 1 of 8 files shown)


lonegunmanb:
Hi @carvido1 , thanks for opening this pr, but unfortunately, according to [terraform-azurerm-provider pr #8982](https://github.com/hashicorp/terraform-provider-azurerm/pull/8982), since the [v2.35.0](https://github.com/hashicorp/terraform-provider-azurerm/blob/main/CHANGELOG-v2.md#2350-november-05-2020), the `node_taints` argument is no longer be configured.

carvido1:
Hello @lonegunmanb .
According to the official azurerm documentation for the resource `azurerm_kubernetes_cluster_node_pool` the [node_tains attribute](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster_node_pool#node_taints) is available.
The `node_taints` was removed from the resource `azurerm_kubernetes_cluster` because it was implemented at node_pool level. This happened when AKS started using VMSS instead of the old nodepools.
Please, reconsider the PR as it can be used to follow some best practices that Microsoft recommends.
Thanks in advance.

lonegunmanb:
Hi @carvido1 , thanks for clarifying. According to the provider's [code](https://github.com/hashicorp/terraform-provider-azurerm/blob/main/internal/services/containers/kubernetes_nodepool.go#L625):

>The AKS API has removed support for tainting all nodes in the default node pool and it is no longer possible to configure this. To taint a node pool, create a separate one.

So I think we cannot add taints to the default node pool anymore, sorry.

carvido1:
Hello.
I've reviewed the code @lonegunmanb. With a test I could have arrived to the return null ...
Thanks for your review. We can close the PR ;)

PR #169, opened 2022-05-30
Add `os_disk_type` for node pools

### LINKED ISSUE #114: Expose `os_disk_type` as a variable

lebenitza:
Fixes #114 

Changes proposed in the pull request: Finally add `os_disk_type`

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -92,6 +92,12 @@ variable "os_disk_size_gb" {
   default     = 50
 }
 
+variable "os_disk_type" {
+  description = "The type of disk which should be used for the Operating System. Possible values are Ephemeral and Managed. Defaults to Managed. Changing this forces a new resource to be created."
+  type        = string
+  default     = "Managed"
+}
+
 variable "private_cluster_enabled" {
   description = "If true cluster API server will be exposed only on internal IP address and available only in cluster vnet."
   type        = bool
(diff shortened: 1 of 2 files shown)


variables.tf
```diff
@@ -92,6 +92,12 @@ variable "os_disk_size_gb" {
   default     = 50
 }
 
+variable "os_disk_type" {
+  description = "The type of disk which should be used for the Operating System. Possible values are Ephemeral and Managed. Defaults to Managed. Changing this forces a new resource to be created."
```
lonegunmanb, line 96: Could we use the following description to be align with the [document](https://github.com/hashicorp/terraform-provider-azurerm/blob/main/website/docs/r/kubernetes_cluster.html.markdown?plain=1#L349)?

```hcl
description = "The type of disk which should be used for the Operating System. Possible values are `Ephemeral` and `Managed`. Defaults to `Managed`. Changing this forces a new resource to be created."
```

lonegunmanb (changes requested):

PR #172, opened 2022-06-02
workaround for recreate bug

davidkarlsen:

This allows to define `location` for the resources explicitly. Because if you have slight diffs on the resource group (like for instance a label being propagated by a mitigating policy), then terraform will say that the `location` is only known after apply, and this will again trigger a recreate on the cluster - blowing the whole cluster away.

The change is backwards-compatible, as the new parameter is opt-in, and old behaviour of reading location from resource-group is preserved.


lonegunmanb (approved):

lonegunmanb:
Thanks @davidkarlsen , LGTM.

Issue #194, opened 2022-06-21
Make cluster's `disk_encryption_set_id` argument configurable.


lonegunmanb:
When writing pr for #183, I found that current module do not support assign `disk_encryption_set_id` to the aks cluster. We should consider make it configurable so the user can improve cluster's security level.

Issue #183, opened 2022-06-20
Improve the test fixture's security.


lonegunmanb:
We've met some compliance issues when we scan this module with [bridgecrew's Checkov](https://bridgecrew.io/checkov/):

Check: CKV_AZURE_117: "Ensure that AKS uses disk encryption set"
        FAILED for resource: module.aks.azurerm_kubernetes_cluster.main
        File: /main.tf:10-142
        Calling File: /test/fixture/main.tf:33-74
        Guide: https://docs.bridgecrew.io/docs/ensure-that-aks-uses-disk-encryption-set

                Code lines for this resource are too many. Please use IDE of your choice to review the file.
Check: CKV_AZURE_141: "Ensure AKS local admin account is disabled"
        FAILED for resource: module.aks.azurerm_kubernetes_cluster.main
        File: /main.tf:10-142
        Calling File: /test/fixture/main.tf:33-74

                Code lines for this resource are too many. Please use IDE of your choice to review the file.
Check: CKV_AZURE_5: "Ensure RBAC is enabled on AKS clusters"
        FAILED for resource: module.aks.azurerm_kubernetes_cluster.main
        File: /main.tf:10-142
        Calling File: /test/fixture/main.tf:33-74
        Guide: https://docs.bridgecrew.io/docs/bc_azr_kubernetes_2

                Code lines for this resource are too many. Please use IDE of your choice to review the file.
Check: CKV_AZURE_4: "Ensure AKS logging to Azure Monitoring is Configured"
        FAILED for resource: module.aks.azurerm_kubernetes_cluster.main
        File: /main.tf:10-142
        Calling File: /test/fixture/main.tf:33-74
        Guide: https://docs.bridgecrew.io/docs/bc_azr_kubernetes_1

                Code lines for this resource are too many. Please use IDE of your choice to review the file.
Check: CKV_AZURE_117: "Ensure that AKS uses disk encryption set"
        FAILED for resource: module.aks_without_monitor.azurerm_kubernetes_cluster.main
        File: /main.tf:10-142
        Calling File: /test/fixture/main.tf:76-84
        Guide: https://docs.bridgecrew.io/docs/ensure-that-aks-uses-disk-encryption-set

                Code lines for this resource are too many. Please use IDE of your choice to review the file.
Check: CKV_AZURE_115: "Ensure that AKS enables private clusters"
        FAILED for resource: module.aks_without_monitor.azurerm_kubernetes_cluster.main
[...]

PR #195, opened 2022-06-21
Add support for `disk_encryption_set_id`

lonegunmanb:
This patch add support for `disk_encryption_set_id` to harden the cluster's security and solve #194 . We've used Checkov and Tflint to scan the test code and ignore one issue by comment. This patch should partially solve #183.


lonegunmanb:
Thanks @jiaweitao001 , would please give this pr another review? Thanks.

Issue #196, opened 2022-06-22
Some identity issues


lonegunmanb:
The `system_assigned_identity` in output should be renamed since it could be user assigned identity. 
The identity-related variables are lacking proper validation.

PR #197, opened 2022-06-22
Improve identity related variables and output.

lonegunmanb:
Rename `var.system_assigned_identity` to `cluster_identity`, add validation and precondition for identity-related variables. Bump Terraform required version to 1.2.0 since we've used precondition. #196 related.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -152,8 +152,19 @@ resource "azurerm_kubernetes_cluster" "main" {
   oidc_issuer_enabled = var.oidc_issuer_enabled
 
   tags = var.tags
-}
 
+  lifecycle {
+    precondition {
+      condition     = (var.client_id != "" && var.client_secret != "") || (var.identity_type != "")
+      error_message = "Either `client_id` and `client_secret` or `identity_type` must be set."
+    }
+    precondition {
+      # Why don't use var.identity_ids != null && length(var.identity_ids)>0 ? Because bool expression in Terraform is not short circuit so even var.identity_ids is null Terraform will still invoke length function with null and cause error. https://github.com/hashicorp/terraform/issues/24128
+      condition     = (var.client_id != "" && var.client_secret != "") || (var.identity_type == "SystemAssigned") || (var.identity_ids == null ? false :length(var.identity_ids) > 0)
+      error_message = "If use identity and `UserAssigned` or `SystemAssigned, UserAssigned` is set, an `identity_ids` must be set as well."
+    }
+  }
+}
 
 resource "azurerm_log_analytics_workspace" "main" {
   count               = var.enable_log_analytics_workspace ? 1 : 0
--- a/outputs.tf
+++ b/outputs.tf
@@ -55,7 +55,7 @@ output "http_application_routing_zone_name" {
 }
 
 output "cluster_identity" {
-  value = azurerm_kubernetes_cluster.main.identity
+  value = try(azurerm_kubernetes_cluster.main.identity[0], null)
 }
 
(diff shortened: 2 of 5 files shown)


outputs.tf
```diff
@@ -54,8 +54,8 @@ output "http_application_routing_zone_name" {
   value = azurerm_kubernetes_cluster.main.http_application_routing_zone_name != null ? azurerm_kubernetes_cluster.main.http_application_routing_zone_name : ""
 }
 
-output "system_assigned_identity" {
-  value = azurerm_kubernetes_cluster.main.identity
+output "cluster_identity" {
+  value = try(azurerm_kubernetes_cluster.main.identity[0], null)
```
zioproto, line 58: Why the `try` block ? Is there a condition when the cluster has `null` identity ?

main.tf
```diff
@@ -152,8 +152,19 @@ resource "azurerm_kubernetes_cluster" "main" {
   oidc_issuer_enabled = var.oidc_issuer_enabled
 
   tags = var.tags
-}
 
+  lifecycle {
+    precondition {
+      condition     = (var.client_id != "" && var.client_secret != "") || (var.identity_type != "")
```
zioproto, line 158: identity_type should have `nullable = false` in the variable definition ?

variables.tf
```diff
@@ -328,9 +330,14 @@ variable "ingress_application_gateway_subnet_id" {
 }
 
 variable "identity_type" {
```
zioproto, line 332: Also identity_type should be `nullable = false` to make sure the condition in main.tf L158 works all the times.

outputs.tf
zioproto: Why we output in the value only the first element, rather than making available the all list ?

If we export the all list the user of the module can use the output with a index and consume all the identities of the cluster in other resources.

This would be also consistent with how we export the kubelet identity, example:
https://github.com/zioproto/istio-aks-example/blob/main/registry.tf#L10

zioproto (changes requested):

outputs.tf
lonegunmanb: Hi, according to the [resource schema](https://github.com/hashicorp/terraform-provider-azurerm/blob/main/internal/services/containers/kubernetes_cluster_resource.go#L238) and [identity schema it referenced](https://github.com/hashicorp/terraform-provider-azurerm/blob/main/vendor/github.com/hashicorp/go-azure-helpers/resourcemanager/commonschema/identity_system_or_user.go#L19-L24), the identity in aks resource is a list with maximum length 1. So in this case the identity list has three cases: `null`, an empty list and a list with exactly one item.

main.tf
lonegunmanb: `identity_type` has a validation block, `null` won't pass the validation so I think it's ok for this case.

variables.tf
lonegunmanb: `identity_type` has a validation block, `null` won't pass the validation so I think it's ok for this case.

outputs.tf
zioproto: What happens when you use `SystemAssigned, UserAssigned` ? Dont you have 2 cluster identities ?

zioproto: Should we also patch the [CHANGLOG.md](https://github.com/Azure/terraform-azurerm-aks/blob/master/CHANGLOG.md) to indicate that the output name changed, so folks using the module must update their code ?

jiaweitao001 (approved):

Issue #198, opened 2022-06-23
Config cluster in test code to use Azure Active Directory and Kubernetes RBAC


lonegunmanb:
According to [BridgeCrew](https://docs.bridgecrew.io/docs/bc_azr_kubernetes_2)'s policy, we should config the cluster in the test code to use AD and K8s RBAC.

Here's a [document](https://docs.microsoft.com/en-us/azure/aks/azure-ad-rbac) about how to use this feature.

PR #199, opened 2022-06-23
Add arguments to `azure_active_directory_role_based_access_control` block, turn rbac in test code on.

lonegunmanb:
Add arguments to `azure_active_directory_role_based_access_control` block, turn rbac in test code. 

Related to #198 and #183


zioproto (approved):
LGTM

Issue #202, opened 2022-06-24
Uses Azure policies add-on in the test fixture code


lonegunmanb:
According to [Checkov's policy](https://docs.bridgecrew.io/docs/ensure-that-aks-uses-azure-policies-add-on):

>Azure Policy Add-on for Kubernetes service (AKS) extends Gatekeeper v3, an admission controller webhook for Open Policy Agent (OPA), to apply at-scale enforcements and safeguards on your clusters in a centralized, consistent manner.

We should use Azure policies add-on in the test fixture code.

PR #203, opened 2022-06-24
Uses Azure policies add-on in the test fixture code

lonegunmanb:
Rename `var.enable_azure_policy` to `var.azure_policy_enabled` to meet the naming convention.

Set `azure_policy_enabled` to true in test fixture code.

Related to #202 and #183

### DIFF
--- a/test/fixture/main.tf
+++ b/test/fixture/main.tf
@@ -40,7 +40,7 @@ module "aks" {
   vnet_subnet_id                   = azurerm_subnet.test.id
   os_disk_size_gb                  = 60
   enable_http_application_routing  = true
-  enable_azure_policy              = true
+  azure_policy_enabled             = true
   enable_host_encryption           = true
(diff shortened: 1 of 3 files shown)


test/fixture/main.tf
```diff
@@ -40,7 +40,7 @@ module "aks" {
   vnet_subnet_id                  = azurerm_subnet.test.id
   os_disk_size_gb                 = 60
   enable_http_application_routing = true
```
zioproto, line 42: There are many "enable" variables that have the wrong naming convention. Does it makes sense to fix all the variables in a single PR ?

lonegunmanb: Thanks for the tip, I think this is a special case since it's related to is single pr.
I agree with you, we need a single pr to fix all these `enable_xxx` variables.

zioproto:
LGTM

PR #207, opened 2022-06-29
feat(versions.tf): TF version must be >= 1.1.0

the-technat:
Fixes #000 

Changes proposed in the pull request: The `variables.tf` is using a feature introduced with Terraform version `v1.1.0` which is called [Disallowing Null Input Values](https://www.terraform.io/language/values/variables#disallowing-null-input-values). To prevent Errors during a Terraform run, I added this version constraint and run fmt/terraform-docs.

Tried to run the tests and they fail with Terraform version `0.14.4` obviously. But seems like there is no Image with tag `1.1.0`. If using the latest version of the image, Terraform version `0.12.10` is installed...

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,14 +224,14 @@ contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additio
 
 | Name | Version |
 |------|---------|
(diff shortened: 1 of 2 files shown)


lonegunmanb (approved):
Thanks @the-technat , LGTM!

PR #209, opened 2022-07-01
Create tls-private-key only if needed

zioproto:
Create the tls-private-key only if needed to avoid confusion with users not knowing the codebase.
The users might expect to use the private key when it is created, so it should not be created at all if there is no use for it.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -30,7 +30,7 @@ resource "azurerm_kubernetes_cluster" "main" {
 
       ssh_key {
         # remove any new lines using the replace interpolation function
-        key_data = replace(coalesce(var.public_ssh_key, tls_private_key.ssh.public_key_openssh), "\n", "")
+        key_data = replace(coalesce(var.public_ssh_key, tls_private_key.ssh[0].public_key_openssh), "\n", "")
       }
     }
   }
--- a/main.tf
+++ b/main.tf
@@ -8,6 +8,7 @@ moved {
 }
 
 resource "tls_private_key" "ssh" {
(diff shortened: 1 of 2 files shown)


main.tf
```diff
@@ -8,6 +8,7 @@ moved {
 }
 
 resource "tls_private_key" "ssh" {
+  count = var.admin_username == null ? 0 : 1
```
lonegunmanb, line 11: Since we have added `count` to this resource we need add index to the reference to `tls_provate.key.ssh`.

lonegunmanb (changes requested):

main.tf
zioproto: You are right. I added the index reference. I will have to test if this block will work after this change:
```
moved {
  from = module.ssh-key.tls_private_key.ssh
  to   = tls_private_key.ssh
}
```

Would be a valid

lonegunmanb: I think it's a tricky scenario since the original code declare `tls_private_key` resource unconditionally, what if a user want to migrate the existing state to new version meanwhile `var.admin_username` is `null`? We might need an upgrade test to see what'll happen in this circumstance.

lonegunmanb, line 11: We have a format issue here, the space is no aligned, would you please run `terraform format` on this file? Thanks!

lonegunmanb (changes requested):

main.tf
zioproto: Done

lonegunmanb (approved):
Thanks @zioproto , LGTM.

PR #213, opened 2022-07-06
feat: Support existing log analytics workspace (#211)

### LINKED ISSUE #211: `enable_log_analytics_workspace` cannot use existing Log Analytics Workspace
I think it would be helpful to use `enable_log_analytics_workspace` being able to pass an existing `log_analytics_workspace_id`.

viters:
Fixes #211 

Provides a way to attach existing Log Analytics Workspace to AKS through Container Insights (`azurerm_log_analytics_solution`).


main.tf
```diff
@@ -171,8 +171,8 @@ resource "azurerm_log_analytics_solution" "main" {
   solution_name         = "ContainerInsights"
   location              = coalesce(var.location, data.azurerm_resource_group.main.location)
   resource_group_name   = var.resource_group_name
-  workspace_resource_id = azurerm_log_analytics_workspace.main[0].id
-  workspace_name        = azurerm_log_analytics_workspace.main[0].name
+  workspace_resource_id = var.log_analytics_workspace == null ? azurerm_log_analytics_workspace.main[0].id : var.log_analytics_workspace.id
```
lonegunmanb, line 174: Hi, do we need add this `var.log_analytics_workspace == null` criteria to `azurerm_log_analytics_solution`'s `count` expression, since the log analytics resource is injected by the module caller？

lonegunmanb (changes requested):
Hello @viters , thanks for submitting this pr. Almost LGTM but only one question.

main.tf
viters: `azurerm_log_analytics_solution` is `ContainerInsights` solution connecting newly created AKS and existing (or just created) Workspace. I think that current `count` expression `var.enable_log_analytics_workspace ? 1 : 0` is valid. If caller specifies `enable_log_analytics_workspace = true` then caller either wants to create a new Workspace or use existing. In other words `ContainerInsights` solution is created always when `enable_log_analytics_workspace = true`, but the created solution can be connected to different Workspaces, which is solved later by conditions on `workspace_resource_id` and `workspace_name`.

viters: I have a question though, because I checked in `azurerm_log_analytics_solution` docs that:
image

In my current solution, existing Workspace `var.log_analytics_workspace` must be placed in the same RG as `var.resource_group_name`. Do you think it is valid assumption? It was valid in my use case.

lonegunmanb: > `azurerm_log_analytics_solution` is `ContainerInsights` solution connecting newly created AKS and existing (or just created) Workspace. I think that current `count` expression `var.enable_log_analytics_workspace ? 1 : 0` is valid. If caller specifies `enable_log_analytics_workspace = true` then caller either wants to create a new Workspace or use existing. In other words `ContainerInsights` solution is created always when `enable_log_analytics_workspace = true`, but the created solution can be connected to different Workspaces, which is solved later by conditions on `workspace_resource_id` and `workspace_name`.

My concern is what if the caller has used another module that creates log analytics domain resources? It looks like we need both workspace and solution to make log work, so maybe we should treat them as an inseparable entity?

lonegunmanb: I've verified that the workspace can be put in different rg, but must be in the same rg with solution, so I prefer we add this `var.log_analytics_workspace == null` criteria to solution's `count` expression, leave the choice to the caller.

lonegunmanb: Hi @viters my concern is can our module be composited with other modules easily. When you opened this pr I do agree with you because what if the caller want to use another module to manage log analytics related resources? Your pr makes it possible to avoid a duplicated workspace and leaves the choice to the caller.

So what if the caller has a module that manage all log analytics related resources, even in a standalone repo along with its own state backend? There's no way for the caller to avoid bind the workspace to a solution that he doesn't want.

I think you pr's value is giving callers the right of composition, so can we give the caller more? Thanks!

viters: @lonegunmanb 
Oh, I got it. Your idea is to add a way to skip creation of `azurerm_log_analytics_solution` as well! I understand now, I will get to it!

main.tf
```diff
@@ -156,10 +156,10 @@ resource "azurerm_kubernetes_cluster" "main" {
 
 
 resource "azurerm_log_analytics_workspace" "main" {
-  count               = var.enable_log_analytics_workspace ? 1 : 0
+  count               = var.enable_log_analytics_workspace && var.log_analytics_workspace == null ? 1 : 0
   name                = var.cluster_log_analytics_workspace_name == null ? "${var.prefix}-workspace" : var.cluster_log_analytics_workspace_name
   location            = coalesce(var.location, data.azurerm_resource_group.main.location)
-  resource_group_name = var.resource_group_name
+  resource_group_name = var.log_analytics_workspace_resource_group_name != null ? var.log_analytics_workspace_resource_group_name : var.resource_group_name
```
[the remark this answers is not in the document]
viters: Yes, I will check how `coalesce` works and update!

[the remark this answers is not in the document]
viters: @lonegunmanb Used `coalesce` as you suggested!

main.tf
```diff
@@ -170,9 +170,9 @@ resource "azurerm_log_analytics_solution" "main" {
   count                 = var.enable_log_analytics_workspace ? 1 : 0
   solution_name         = "ContainerInsights"
   location              = coalesce(var.location, data.azurerm_resource_group.main.location)
-  resource_group_name   = var.resource_group_name
-  workspace_resource_id = azurerm_log_analytics_workspace.main[0].id
-  workspace_name        = azurerm_log_analytics_workspace.main[0].name
+  resource_group_name   = var.log_analytics_workspace_resource_group_name != null ? var.log_analytics_workspace_resource_group_name : var.resource_group_name
```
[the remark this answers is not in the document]
viters: @lonegunmanb Used `coalesce` as you suggested!

main.tf
viters: @lonegunmanb 
As we discussed, I've added `var.log_analytics_solution_id` and `azurerm_log_analytics_solution` resource will not be created if `var.log_analytics_solution_id == null`.

I've used `string id` instead of `bool` to avoid breaking changes in future - we do not need the solution ID now, but it may be required in future Azure API somewhere, e.g. in `oms_agent` block.

lonegunmanb (approved):
Thanks @viters, LGTM!

PR #218, opened 2022-07-11
Make `local_account_disabled` configurable to fix #216.

### LINKED ISSUE #216: We need make `local_account_disabled` argument configurable.
As [BridgeCrew Checkov's policy](https://github.com/bridgecrewio/checkov/blob/ea6cf4e82919c5587a90e90e0e0224fcc76016d9/checkov/terraform/checks/resource/azure/AKSLocalAdminDisabled.py) suggested, we should ensure AKS local admin account is disabled, so we need make `local_account_disabled` argument configurable.

Related to #183.

lonegunmanb:
Make `local_account_disabled` configurable to fix #216.


jiaweitao001 (approved):
Ship it.

PR #219, opened 2022-07-11
Enable private cluster for the test code.

### LINKED ISSUE #217: Turn on `private_cluster_enabled` for test code.
According to [BridgeCrew Checkov's policy](https://docs.bridgecrew.io/docs/ensure-that-aks-enables-private-clusters), we should ensure AKS enables private clusters. 

Related to #183

lonegunmanb:
According to [BridgeCrew Checkov's policy](https://docs.bridgecrew.io/docs/ensure-that-aks-enables-private-clusters), we should ensure AKS enables private clusters. 

Related to #183 and fix #217.


jiaweitao001 (approved):
Change LGTM.

Issue #108, opened 2021-06-23
Add parameter api_server_authorized_ip_ranges


fplantinga-guida:
Add this feature to allow authorized ip ranges configuration.

lonegunmanb:
I'm closing this issue because this feature has been implemented by #220 . Thanks @fplantinga-guida !

PR #220, opened 2022-07-11
Make `api_server_authorized_ip_ranges` configurable.

### LINKED ISSUE #208: Make `api_server_authorized_ip_ranges` argument configurable. Assign `api_server_authorized_ip_ranges` in test code to meet the compliance.
According to [Checkov's policy](https://docs.bridgecrew.io/docs/bc_azr_kubernetes_3):

>To enhance cluster security and minimize attacks, the API server should only be accessible from a limited set of IP address ranges. These IP ranges allow defined IP address ranges to communicate with the API server. A request made to the API server from an IP address that is not part of these authorized IP ranges is blocked.

We need add variable for `api_server_authorized_ip_ranges` and set this variable in the test code to make sure our test code meet the compliance.

lonegunmanb:
Make `api_server_authorized_ip_ranges` configurable.

Fix #208 and #108

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -26,6 +26,7 @@ resource "azurerm_kubernetes_cluster" "main" {
   private_dns_zone_id                 = var.private_dns_zone_id
   private_cluster_public_fqdn_enabled = var.private_cluster_public_fqdn_enabled
   local_account_disabled              = var.local_account_disabled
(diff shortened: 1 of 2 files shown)


jiaweitao001:
Looks good to me. I think it's better to include it in the test `main.tf` as well so the users can have a good example.

PR #226, opened 2022-07-14
Add index to `to` address to fix #225

### LINKED ISSUE #225: The `to` address inside the `moved` block is incorrect.
As now the `tls_private_key` resource has a `count` meta-argument through #209, `to` address inside the `moved` block is now lacking of an index. It should be:

```hcl
moved {
  from = module.ssh-key.tls_private_key.ssh
  to   = tls_private_key.ssh[0]
}
```

lonegunmanb:
Add index to `to` address to fix #225

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -4,7 +4,7 @@ data "azurerm_resource_group" "main" {
 
 moved {
   from = module.ssh-key.tls_private_key.ssh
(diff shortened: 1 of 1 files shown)


zioproto:
LGTM !
I missed this when I added the `count` in PR #209

PR #229, opened 2022-08-15
Loose the restriction on tls provider's version to include major version greater than 3.0

### LINKED ISSUE #228: bug: backwards incompatible TLS provider restriction
### Description

After the v5 release, a TLS provider restriction has been introduced to the codebase, namely TLS provoder version `~> 3.1`.
See: https://github.com/Azure/terraform-azurerm-aks/blob/5.0.0/versions.tf

Please note that there is TLS provider version 4.x already available for a while now, and there are users of this AKS module who are potentially already using AKS module v4 (even in production) with the newer TLS provider (v4).  Since there is a restriction in Terraform, that the older TLS provider (v3) is not able to manage the resource managed by the newer TLS provider (v4) the Terraform workspaces in this situation are unable to run to completion (failing at initialization).

The AKS module's version 4 did not have any restrictions on the TLS provider version:
https://github.com/Azure/terraform-azurerm-aks/blob/4.16.0/versions.tf

Even if it is best practice to have version constraints, I recommend also including v4 TLS provider as an option, since there is not much change in the key resource being managed by the provider code.

Thanks.

lonegunmanb:
This pr should fix #228 issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -347,14 +347,14 @@ contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additio
 |---------------------------------------------------------------------------|---------|
 | <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 1.2  |
 | <a name="requirement_azurerm"></a> [azurerm](#requirement\_azurerm)       | ~> 3.3  |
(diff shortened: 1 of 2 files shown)


jiaweitao001 (approved):
LGTM. Thanks!

PR #241, opened 2022-09-05
Prepare v6.0 and new CI pipeline.

lonegunmanb:
This is a major version upgrade and it mainly focus on adding a new CI pipeline to this repo so we can speed up our code review and enforce a high code quality standard.

It also split original test fixture folder into three so we can demonstrate different use cases.


zioproto:
Do we actually need `az` in the container ? Now it is not there:

```
% docker run -ti --rm -v $(pwd):/src -w /src mcr.microsoft.com/azterraform:latest /bin/bash
root@53188d41e7b6:/src# az
bash: az: command not found
```

zioproto (approved):
LGTM

PR #242, opened 2022-09-09
Adds support for Ultra disks

digiserg:
Adds support for Ultra Disks by enabling the option 

[ultra_ssd_enabled](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster#ultra_ssd_enabled) - (Optional) Used to specify whether the UltraSSD is enabled in the Default Node Pool. Defaults to false. See [the documentation](https://docs.microsoft.com/azure/aks/use-ultra-disks) for more information.


lonegunmanb:
Hi @digiserg , thanks for opening this pr! We've just released v6.0.0 and in this new major release we've added a new CI pipeline. Would you please update your pr with the latest master branch so we can leverage our CI pipeline to do the check and tests for us? We also provided a docker image so we can run the code check and tests on our local machine, appreciate if we can read the latest README and do the check first, thanks!

digiserg:
I'll send a new clean PR, this has become a mess

PR #243, opened 2022-09-12
Add outputs for created Log Analytics workspace

zioproto:
Add outputs for created Log Analytics workspace


lonegunmanb (approved):
Thanks @zioproto, LGTM

PR #262, opened 2022-09-28
Improve placeholders for visibility in the UX

zioproto:
I was opening a bug and the green "Submit new issue" button was disabled because of a missing required value. I spent a lot of time understanding what the problem was because I could not distinguish the placeholder from the actual value:

### Placeholder:
Screenshot 2022-09-28 at 09 28 49

### Actual value typed in:

Screenshot 2022-09-28 at 09 28 42

This patch improves the placeholders so that it is more visibile if the user forgot to enter actual data.


lonegunmanb (approved):
LTGM, thanks @zioproto

PR #265, opened 2022-10-10
[Breaking]  Wrap `log_analytics_solution_id` to an object to fix #263.

### LINKED ISSUE #263: Wraping `var.log_analytics_solution_id` with an object so we can create an `azurerm_log_analytics_solution` outside this module and pass id as a variable.

### Description

Now we're using `count = var.log_analytics_workspace_enabled && var.log_analytics_solution_id == null ? 1 : 0` to determine whether we should create an `azurerm_log_analytics_solution` or not, that requires the `var.log_analytics_solution_id` is a known value during `terraform plan`. If we'd like to create a custom solution and pass it's id via this variable, it'll raise an error. It'll be nice if we can wrap this `var.log_analytics_solution_id` with an `object`, just like what we've done with `var.log_analytics_workspace`.

`azurerm_log_analytics_solution`, `var.log_analytics_solution_id`

```hcl
variable "log_analytics_solution" {
  type = object({
    id = string
  })
  description = "(Optional) Existing azurerm_log_analytics_solution ID. Providing ID disables creation of azurerm_log_analytics_solution."
  default     = null
  nullable    = true
  validation {
    condition = var.log_analytics_solution == null ? true : (var.log_analytics_solution.id != null)
    error_message = "`var.log_analytics_solution.id` is required when `var.log_analytics_solution` is not `null`."
  }
}
```

### References

Due to [Terraform's binary operator is not short-circuit](https://github.com/hashicorp/terraform/issues/24128) I'm using:

```hcl
condition = var.log_analytics_solution == null ? true : (var.log_analytics_solution.id != null)
```

instead of:

```hcl
condition = var.log_analytics_solution == null || var.log_analytics_solution.id != null
```

lonegunmanb:
Wrap `log_analytics_solution_id` to an object to fix #263.
This pr also added symbolic links notice in the readme.


jiaweitao001 (approved):
LGTM. 🚢

PR #212, opened 2022-07-06
Possibility to setup multiple node pools

Sanverik:
Changes proposed in the pull request:
- extend functionality to be able to specify multiple node pools within the module

PR #267, opened 2022-10-14
Additional node pools submodule

nlamirault:
Code from : https://github.com/Azure/terraform-azurerm-aks/pull/212

### DIFF
--- a/modules/node-pools/data.tf
+++ b/modules/node-pools/data.tf
@@ -1,0 +1,4 @@
+data "azurerm_kubernetes_cluster" "main" {
+  name                = var.cluster_name
+  resource_group_name = var.resource_group_name
+}
--- a/modules/node-pools/main.tf
+++ b/modules/node-pools/main.tf
@@ -1,0 +1,37 @@
+resource "azurerm_kubernetes_cluster_node_pool" "workers" {
+  for_each = var.additional_node_pools
+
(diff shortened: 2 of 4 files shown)


lonegunmanb:
Thansk @nlamirault for opening this pr. Personally I would prefer not to add additional node pool in this module since I'd like to maintain this module only focus on the Aks control plane itself. 

This pr has high quality, I'd like to create a standalone module for it rather than making it a submodule nested in this aks module. I was working on a [template repo](https://github.com/lonegunmanb/terraform-verified-module) but I don't have enough time to polish it. I'll be happy create a new module based on this template and your pr, or you can create your own one, then we can see what we could do to publish it as a new Azure module, sounds good?

nlamirault:
@lonegunmanb i will wait for your new repository and after that, i will submit a new PR to it.

lonegunmanb:
An update, I've changed my mind. There're only two resources types for Aks: `azurerm_kubernetes_cluster` and `azurerm_kubernetes_cluster_node_pool`, we cannot create a new module which only contains node pool, so this module should support node pool.

We'll implement this feature in our next release, maybe we can have it in April.

lonegunmanb:
I'm closing this pr because the feature has been implemented in #323. Thanks for your contribution @nlamirault !

Issue #137, opened 2022-01-13
Allow RBAC for K8s Service Accounts with out the need for AzureAD 


manoatsas:
Allow Kubernetes RBAC to be enabled to use Kubernetes Service Accounts with out the need for AzureAD integration per docs
- Azurerm TF provider - https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster#role_based_access_control, `azure_active_directory` is optional for `role_based_access_control` block
- Azure docs - https://docs.microsoft.com/en-us/azure/aks/concepts-identity#azure-role-based-access-control 

However, when `enable_role_based_access_control = true` and with default `rbac_aad_managed=false` currently it fails with the error
```
│ Error: You must specify client_app_id and server_app_id and server_app_secret when using managed aad rbac (managed = false)
│ 
│   with module.aks.azurerm_kubernetes_cluster.main,
│   on .terraform/modules/aks/main.tf line 10, in resource "azurerm_kubernetes_cluster" "main":
│   10: resource "azurerm_kubernetes_cluster" "main" {
```

zioproto:
The `role_based_access_control` block is deprecated in favour of the `role_based_access_control_enabled` property and the `azure_active_directory_role_based_access_control` block.

@manoatsas I struggle to understand the configuration you would like to see in the module.

Can you provide a working example of the resulting `azurerm_kubernetes_cluster` resource that would like the module to produce ?

thanks

lonegunmanb:
I'm closing this issue since no response from the user, please feel free to reopen it if you have any further question @manoatsas.

PR #269, opened 2022-10-18
Azure AD RBAC enable/disable with variable rbac_aad

### LINKED ISSUE #215: Can't use role_based_access_control without azure_active_directory
Hi, in the lates version - 4.16.0, there is no ability to deploy AKS with only RBAC and without AAD, there is no option to disable "AKS-managed AAD"  only if also disable RBAC
It will always create "azure_active_directory" even "rbac_aad_managed = false" 
```
    dynamic "azure_active_directory" {
      for_each = var.enable_role_based_access_control && var.rbac_aad_managed ? ["rbac"] : []
      content {
        managed                = true
        admin_group_object_ids = var.rbac_aad_admin_group_object_ids
      }
    }

    dynamic "azure_active_directory" {
      for_each = var.enable_role_based_access_control && !var.rbac_aad_managed ? ["rbac"] : []
      content {
        managed           = false
        client_app_id     = var.rbac_aad_client_app_id
        server_app_id     = var.rbac_aad_server_app_id
        server_app_secret = var.rbac_aad_server_app_secret
      }
    }
```

zioproto:
Fixes #215 

The current code was using the variable `var.role_based_access_control_enabled` to check if creating the dynamic block `azure_active_directory_role_based_access_control`. 

This is wrong because there is no way to use just Kubernetes RBAC without enabling the dynamic block `azure_active_directory_role_based_access_control`.


lonegunmanb (changes requested):
Thanks @zioproto for opening this pr, unfortunately we have some e2e test failures here.

```text
│ Error: Operation failed
        	            	│ 
        	            	│   on ../../main.tf line 96, in resource "azurerm_kubernetes_cluster" "main":
        	            	│   96:     for_each = var.role_based_access_control_enabled && var.rbac_aad_azure_rbac_enabled && var.rbac_aad_managed ? ["rbac"] : []
        	            	│     ├────────────────
        	            	│     │ var.rbac_aad_azure_rbac_enabled is null
        	            	│     │ var.role_based_access_control_enabled is true
        	            	│ 
        	            	│ Error during operation: argument must not be null.
```

Would you please fix the error and let's try again? Thanks!

main.tf
```diff
@@ -95,7 +95,7 @@ resource "azurerm_kubernetes_cluster" "main" {
     }
   }
   dynamic "azure_active_directory_role_based_access_control" {
-    for_each = var.role_based_access_control_enabled && var.rbac_aad_managed ? ["rbac"] : []
+    for_each = var.role_based_access_control_enabled && var.rbac_aad_azure_rbac_enabled && var.rbac_aad_managed ? ["rbac"] : []
```
lonegunmanb, line 98: The only way that we can set an `azure_active_directory_role_based_access_control` block with `var.rbac_aad_managed = true` is the `azure_active_directory_role_based_access_control.azure_rbac_enabled = true`, we cannot set it to `null` or `false`, is that correct? Why the provider allows us to configure `azure_rbac_enabled`'s value?

lonegunmanb (commented):
Thanks @zioproto, I have a question for your pr.

lonegunmanb:
Hi @zioproto we got a failed test when we tried to provison `example/startup`:

```hcl
TestExamplesStartup 2022-11-16T01:00:17Z logger.go:66: module.aks.azurerm_kubernetes_cluster.main: Refreshing state... [id=/subscriptions/xxxx/resourceGroups/eb0a5e2abd042c6f-rg/providers/Microsoft.ContainerService/managedClusters/prefix-eb0a5e2abd042c6f-aks]
TestExamplesStartup 2022-11-16T01:00:19Z logger.go:66: ╷
TestExamplesStartup 2022-11-16T01:00:19Z logger.go:66: │ Warning: Argument is deprecated
TestExamplesStartup 2022-11-16T01:00:19Z logger.go:66: │ 
TestExamplesStartup 2022-11-16T01:00:19Z logger.go:66: │   with azurerm_subnet.test,
TestExamplesStartup 2022-11-16T01:00:19Z logger.go:66: │   on main.tf line 31, in resource "azurerm_subnet" "test":
TestExamplesStartup 2022-11-16T01:00:19Z logger.go:66: │   31:   enforce_private_link_endpoint_network_policies = true
TestExamplesStartup 2022-11-16T01:00:19Z logger.go:66: │ 
TestExamplesStartup 2022-11-16T01:00:19Z logger.go:66: │ `enforce_private_link_endpoint_network_policies` will be removed in favour
TestExamplesStartup 2022-11-16T01:00:19Z logger.go:66: │ of the property `private_endpoint_network_policies_enabled` in version 4.0
TestExamplesStartup 2022-11-16T01:00:19Z logger.go:66: │ of the AzureRM Provider
TestExamplesStartup 2022-11-16T01:00:19Z logger.go:66: ╵
TestExamplesStartup 2022-11-16T01:00:19Z logger.go:66: ╷
TestExamplesStartup 2022-11-16T01:00:19Z logger.go:66: │ Error: retrieving Access Profile for Managed Cluster (Subscription: "xxxx"
TestExamplesStartup 2022-11-16T01:00:19Z logger.go:66: │ Resource Group Name: "eb0a5e2abd042c6f-rg"
TestExamplesStartup 2022-11-16T01:00:19Z logger.go:66: │ Resource Name: "prefix-eb0a5e2abd042c6f-aks"): managedclusters.ManagedClustersClient#GetAccessProfile: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="BadRequest" Message="Getting static credential is not allowed because this cluster is set to disable local accounts."
TestExamplesStartup 2022-11-16T01:00:19Z logger.go:66: │ 
TestExamplesStartup 2022-11-16T01:00:19Z logger.go:66: │   with module.aks.azurerm_kubernetes_cluster.main,
TestExamplesStartup 2022-11-16T01:00:19Z logger.go:66: │   on ../../main.tf line 17, in resource "azurerm_kubernetes_cluster" "main":
TestExamplesStartup 2022-11-16T01:00:19Z logger.go:66: │   17: resource "azurerm_kubernetes_cluster" "main" {
TestExamplesStartup 2022-11-16T01:00:19Z logger.go:66: │ 
TestExamplesStartup 2022-11-16T01:00:19Z logger.go:66: ╵
```

zioproto:
The test fails because: `local_account_disabled                  = true`
https://github.com/Azure/terraform-azurerm-aks/blob/8b90d95e468cde5be31d11e4354c75b0dae98e62/examples/startup/main.tf#L56

the only change of previous behavior in this PR is here:

https://github.com/Azure/terraform-azurerm-aks/pull/269/files#diff-dc46acf24afd63ef8c556b77c126ccc6e578bc87e3aa09a931f33d9bf2532fbbL102

The value of `azure_rbac_enabled` is `false` instead of before it was `null`.

I understand that `azure_rbac_enabled` set to `null` has the consequence that `local_account_disabled` is ignored. 

This is probably a different bug, but I have to do more research to see if it comes from the Terraform provider or from AKS itself.

zioproto:
There was a bug in my conditional logic so that the block `azure_active_directory_role_based_access_control` was never created, and for this reason the end to end test was failing.

I pushed now a new commit that introduces a new variable, the `rbac_aad`, in addition to `rbac_aad_managed`.

There was a misconception on my side, where my initial understand was that `rbac_aad_managed` was used to enable/disable the Azure AD integration. This is *wrong*.

The existence of the block `azure_active_directory_role_based_access_control` determines if the Azure AD integration is active or not.

The `managed = true` or `managed = false` inside the block have the following meaning:

* true -> https://learn.microsoft.com/en-us/azure/aks/managed-aad
* false -> https://learn.microsoft.com/en-us/azure/aks/azure-ad-integration-cli

@lonegunmanb please review now

Extra info:  When the block `azure_active_directory_role_based_access_control` is missing completely (that was never tested before) it seems is still possible to give the directive --disable-local-accounts. However this will change in AKS 1.25+

variables.tf
```diff
@@ -409,6 +409,13 @@ variable "public_ssh_key" {
   default     = ""
 }
 
+variable "rbac_aad" {
+  type        = bool
+  description = "(Optional) Is Azure Active Directory ingration enabled?"
+  default     = false
```
zioproto, line 415: I dont like the idea of enabling this by default, so I have set this to `false`.
But this is a breaking change, because this will disable the `azure_active_directory_role_based_access_control` block.

I am in doubt if `true` is a better default to avoid breaking changes.

Cc:
@lonegunmanb

zioproto: As agreed in our meeting, we will set the default value to true because it is the most secure value for the users.

DavidSpek:
Also related too https://github.com/Azure/terraform-azurerm-aks/issues/137

locals.tf
```diff
@@ -18,4 +18,6 @@ locals {
       name = var.log_analytics_workspace.name
     }
   ) : null # Finally, the Log Analytics Workspace should be disabled.
-}
\ No newline at end of file
+  # If var.rbac_aad_azure_rbac_enabled  is null set the local to false. Otherise set the local to the var true/false value
+  rbac_aad_azure_rbac_enabled = var.rbac_aad_azure_rbac_enabled == null ? false : var.rbac_aad_azure_rbac_enabled
```
lonegunmanb, line 22: Could we simplify this expression?:

```hcl
rbac_aad_azure_rbac_enabled = var.rbac_aad_azure_rbac_enabled == true
```

main.tf
```diff
@@ -95,17 +95,17 @@ resource "azurerm_kubernetes_cluster" "main" {
     }
   }
   dynamic "azure_active_directory_role_based_access_control" {
-    for_each = var.role_based_access_control_enabled && var.rbac_aad_managed ? ["rbac"] : []
+    for_each = var.rbac_aad && var.rbac_aad_managed ? ["rbac"] : []
 
     content {
       admin_group_object_ids = var.rbac_aad_admin_group_object_ids
-      azure_rbac_enabled     = var.rbac_aad_azure_rbac_enabled
+      azure_rbac_enabled     = local.rbac_aad_azure_rbac_enabled
```
lonegunmanb, line 102: What if we declare an aks resource with an `azure_active_directory_role_based_access_control` block, but omit it's `azure_rbac_enabled` argument? I mean, if we set `null` to this argument, is that a valid configuration?

lonegunmanb (changes requested):
Thanks @zioproto for this update, almost LGTM but I have one change request and one question.

main.tf
zioproto: It is a valid configuration. In the `master` branch the variable `rbac_aad_azure_rbac_enabled` has already a default value of `null`. This is because [azure_rbac_enabled](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster#azure_rbac_enabled) is optional.

This is a valid configuration and it means:
* Enable the Active Directory integration for authentication
* Do not use Active Director for RBAC

This is the equivalent of doing `az aks create  --enable-aad` **without** the parameter `--enable-azure-rbac`.

locals.tf
zioproto: Yes we can, I tested:
```
% terraform console
> null == true
false
```
So you expression works and it is simpler. Thanks

main.tf
zioproto: We should avoid having the `local` and use directly the variable then ?

zioproto:
@lonegunmanb could you please enable the E2E test for this PR again ? It think now it is good to test and merge. Thanks !

main.tf
```diff
@@ -256,6 +256,10 @@ resource "azurerm_kubernetes_cluster" "main" {
       condition     = local.automatic_channel_upgrade_check
       error_message = "Either disable automatic upgrades, or only specify up to the minor version when using `automatic_channel_upgrade=patch` or don't specify `kubernetes_version` at all when using `automatic_channel_upgrade=stable|rapid|node-image`. With automatic upgrades `orchestrator_version` must be set to `null`."
     }
+    precondition {
+      condition     = var.role_based_access_control_enabled || (!var.role_based_access_control_enabled && !var.rbac_aad)
```
zioproto, line 260: @lonegunmanb could you please review if this precondition looks good to you ? thanks

This is the goal:

| role_based_access_control_enabled | rbac_aad | condition |
| --- | --- | --- |
|True| True | True |
|True| False | True |
|False| True | False |
|False| False | True |

lonegunmanb: I think we can simplify this condition to:

```hcl
condition     = var.role_based_access_control_enabled || !var.rbac_aad
```

lonegunmanb (approved):
Thanks @zioproto, LGTM! 🚀

PR #271, opened 2022-10-25
Add support for pod subnet

mr-onion-2:
Add support for [pod_subnet_id](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster#pod_subnet_id)


scripts
```diff
@@ -0,0 +1 @@
+tfmod-scaffold/scripts
```
lonegunmanb, line 1: My bad, I should ignore this soft link in .gitignore file. Would you please remove this file from this pr? Thanks!

lonegunmanb (changes requested):

scripts
mr-onion-2: Done, sorry didn't notice that!

lonegunmanb:
@mr-onion-2 sorry for the late reply, our test infrastructure was broken last week, and we've just fixed it. We've changed the test runner's pool name, would you please rebase this pr with the latest master branch so our new test pipeline can run the test? Apology for your inconveniency and thanks for your understanding.

lonegunmanb (approved):
Thanks @mr-onion-2 , LGTM!

PR #278, opened 2022-11-21
feat: allow for configuring auto_scaler_profile

DavidSpek:
## Describe your changes

This PR allows for users to specify the `auto_scaler_profile` settings when autoscaling is enabled. Some of these settings are quite vital too having the autoscaler work nicely.

## Issue number

#000

Thanks for your cooperation!


DavidSpek:
@microsoft-github-policy-service agree

zioproto:
Hello, thanks for your contribution @DavidSpek 

Could you please run locally the `pre-commit` and `pr-check` steps as explained here:

https://github.com/Azure/terraform-azurerm-aks#pre-commit--pr-check--test

I see this step is failing for this PR:
```
docker pull mcr.microsoft.com/azterraform:latest 
docker run --rm -v $(pwd):/src -w /src mcr.microsoft.com/azterraform:latest make pr-check
```

the variables in the `variables.tf` file are not in the order expected by the linter.

thanks

DavidSpek:
@zioproto Thanks for the quick response. I had run the commands locally but it seems like it didn't run quite correctly. Everything should be good now.

lonegunmanb (changes requested):
Hello @DavidSpek, thanks for opening this pr! Almost LGTM but we'd like to keep our variables' description and default value as same as the [officical document](https://github.com/hashicorp/terraform-provider-azurerm/blob/main/website/docs/r/kubernetes_cluster.html.markdown?plain=1#L259-L293), so would you please align the description and default value for us? Thanks for your understanding!

variables.tf
```diff
@@ -93,6 +93,109 @@ variable "api_server_authorized_ip_ranges" {
...
+variable "auto_scaler_profile_max_unready_percentage" {
+  description = "The maximum percentage of unready nodes in a node group. Defaults to 45."
+  type        = number
+  default     = 45
+}
+
+variable "auto_scaler_profile_new_pod_scale_up_delay" {
+  description = "The delay after pod starts, before it's added to calculation of pending pods. It should be used together with skip_nodes_with_local_storage. Defaults to 0s."
+  type        = string
+  default     = "0s"
```
lonegunmanb, line 141: Do we have special reason for this `0s` default value since the [provider's default value](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster#new_pod_scale_up_delay) is `10s`?

variables.tf
```diff
@@ -93,6 +93,109 @@ variable "api_server_authorized_ip_ranges" {
...
+variable "auto_scaler_profile_scale_down_unready" {
+  description = "The time after which a node should be deleted after it has been unready for this long. Defaults to 20m."
+  type        = string
+  default     = "20m"
+
+}
+
+variable "auto_scaler_profile_scale_down_utilization_threshold" {
+  description = "The threshold in %!u(MISSING)nder which a node is considered for scale down. Defaults to 0.5."
+  type        = number
```
lonegunmanb, line 177: It looks like the type should be [string](https://github.com/hashicorp/terraform-provider-azurerm/blob/main/internal/services/containers/kubernetes_cluster_resource.go#L201-L205) here.

variables.tf
```diff
@@ -93,6 +93,109 @@ variable "api_server_authorized_ip_ranges" {
...
+
+variable "auto_scaler_profile_expander" {
+  description = "The expander for the node group. Possible values are `random`, `most-pods`, `least-waste`, `price`, `priority`. Defaults to `random`."
+  type        = string
+  default     = "random"
+}
+
+variable "auto_scaler_profile_max_graceful_termination_sec" {
+  description = "The maximum time in seconds for which a node can be gracefully terminated. Defaults to 600."
+  type        = number
```
lonegunmanb, line 116: It looks like the type should be [string](https://github.com/hashicorp/terraform-provider-azurerm/blob/main/internal/services/containers/kubernetes_cluster_resource.go#L136-L140) here.

variables.tf
DavidSpek: I believe the provider default value might not be correct in the documentation since setting it too `10s` showed a change from `0s` too `10s` when running `terraform plan`. Also, the default of the cluster autoscaler itself is `0s`, which further supports my suspicion that the provider documentation is incorrect.

variables.tf
```diff
@@ -93,6 +93,109 @@ variable "api_server_authorized_ip_ranges" {
...
+}
+
+variable "auto_scaler_profile_empty_bulk_delete_max" {
+  description = "The maximum number of empty nodes that can be deleted at the same time. Defaults to 10."
+  type        = number
+  default     = 10
+}
+
+variable "auto_scaler_profile_expander" {
+  description = "The expander for the node group. Possible values are `random`, `most-pods`, `least-waste`, `price`, `priority`. Defaults to `random`."
```
lonegunmanb, line 109: According to the [provider's schema](https://github.com/hashicorp/terraform-provider-azurerm/blob/307e6478cf507c41301413475d77574618031db5/internal/services/containers/kubernetes_cluster_resource.go#L125-L135), the valid values for this argument should be: `least-waste`, `most-pods`, `priority` and `random`.

variables.tf
lonegunmanb: I've tried the following `auto-scaler_profile` block and the config was idempotent:

```hcl
auto_scaler_profile {
    new_pod_scale_up_delay            = "10s"
  }
```

Would you please give us a minimum sample code that `10s` as default value would cause a drift in the generated plan? Thanks!

lonegunmanb (changes requested):
Thanks @DavidSpek for your reply, we still need enum and number values in the description to be correctly backquoted, would you please copy the descriptions from this [markdown document](https://github.com/hashicorp/terraform-provider-azurerm/blob/f8e47e77eaa8baf00918d6172edefe97fdfaaa7b/website/docs/r/kubernetes_cluster.html.markdown?plain=1)? Thanks for your understanding!

DavidSpek:
@lonegunmanb Sorry for the late response. I've made the changes you've requested. Regarding the `new_pod_scale_up_delay`, it might be related to my specific environment. But given that it'll take me quite some time to try and validate that I've changed the PR to be inline with the provider documentation. We can change this to `0s` (which I think it should be set to be default) down the road with a fix in the provider as well.

variables.tf
```diff
@@ -93,6 +93,113 @@ variable "api_server_authorized_ip_ranges" {
...
+variable "auto_scaler_profile_scale_down_delay_after_add" {
+  description = "The delay after node addition, before another node can be removed. Defaults to 10m."
+  type        = string
+  default     = "10m"
+}
+
+variable "auto_scaler_profile_scale_down_delay_after_delete" {
+  description = "The delay after node deletion, before another node can be removed. Defaults to 10s."
+  type        = string
+  default     = "10s"
```
lonegunmanb, line 157: According to the [document](https://github.com/hashicorp/terraform-provider-azurerm/blob/main/website/docs/r/kubernetes_cluster.html.markdown?plain=1#L277):

>Defaults to the value used for `scan_interval`.

Which means if the user omit this value or set it to `null`, we should use `var.auto_scaler_profile_scan_interval` instead. I would suggest a new `local` in `locals.tf` file to implement this logic, and add a new unit test to ensure this logic. I would suggest you to read the `test/unit` folder to see how to write a unit test to do that, or you can just finish all other modifications then ping me, I could compose a new pr to your branch which contains the unit test.

variables.tf
```diff
@@ -93,6 +93,113 @@ variable "api_server_authorized_ip_ranges" {
...
+variable "auto_scaler_profile_scale_down_unready" {
+  description = "The time after which a node should be deleted after it has been unready for this long. Defaults to 20m."
+  type        = string
+  default     = "20m"
+
+}
+
+variable "auto_scaler_profile_scale_down_utilization_threshold" {
+  description = "The threshold in %!u(MISSING)nder which a node is considered for scale down. Defaults to 0.5."
+  type        = number
```
lonegunmanb, line 181: According to the [document](https://github.com/hashicorp/terraform-provider-azurerm/blob/main/internal/services/containers/kubernetes_cluster_resource.go#L201-L205) I would suggest that we set the type to `string` here along with the default value in `string`.

lonegunmanb (changes requested):
Thanks @DavidSpek for this update, we use terraform-docs to generate readme file from terraform code including this `variables.tf`, so it's very important to keep variables' `description` value to valid markdown format so the generated readme file's format could be correct. In the `description` we should backquote the value. I would suggest that we just copy the [provider's document](https://github.com/hashicorp/terraform-provider-azurerm/blob/main/website/docs/r/kubernetes_cluster.html.markdown?plain=1#L259-L293) in provider's repo.

DavidSpek:
@lonegunmanb Sorry for all the back and forth with all this. I just pushed a commit that I think resolves all your comments.

lonegunmanb (changes requested):

lonegunmanb (changes requested):
Hi @DavidSpek thanks for the update, almost LGTM and we've passed the e2e test. Only one change request, we can remove `default = null`, after that, we're good to merge.

Thanks for your understanding and contribution!

DavidSpek:
@lonegunmanb Sorry for the late response, I was away on vacation. I'll remove the merge conflicts and make the last change shortly.

DavidSpek:
@lonegunmanb If I remove the `default = null` for `auto_scaler_profile_scale_down_delay_after_delete` it becomes a required value for the module and the test will fail. I don't think we would want that to be a required value right?

DavidSpek:
@lonegunmanb It seems like `default = null` is now required and used in lots of places. I've added where needed so the tests pass properly.

variables.tf
```diff
@@ -93,6 +93,113 @@ variable "api_server_authorized_ip_ranges" {
...
+  description = "How long a node should be unneeded before it is eligible for scale down. Defaults to `10m`."
+  type        = string
+  default     = "10m"
+}
+
+variable "auto_scaler_profile_scale_down_unready" {
+  description = "How long an unready node should be unneeded before it is eligible for scale down. Defaults to `20m`."
+  type        = string
+  default     = "20m"
+
```
zioproto, line 176: neat: empty line

DavidSpek: done

lonegunmanb:
Hello @DavidSpek , I'm sorry we've met an issue on our pipeline so we have to upgrade the Github action yaml files in this repo, would you please merge your branch with the latest master branch so we can run the e2e test? Apology for the inconvenience, we're still refactoring the pipeline.

DavidSpek:
@lonegunmanb No problem. I've rebased on the latest master and pushed it now.

variables.tf
```diff
@@ -93,6 +93,113 @@ variable "api_server_authorized_ip_ranges" {
...
+variable "auto_scaler_profile_scale_down_delay_after_add" {
+  description = "How long after the scale up of AKS nodes the scale down evaluation resumes. Defaults to `10m`."
+  type        = string
+  default     = "10m"
+}
+
+variable "auto_scaler_profile_scale_down_delay_after_delete" {
+  description = "How long after node deletion that scale down evaluation resumes. Defaults to the value used for `scan_interval`."
+  type        = string
+  default     = null
```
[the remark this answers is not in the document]
lonegunmanb: Oh I've made a silly mistake... LGTM here.

lonegunmanb:
Hello @DavidSpek , we have a version upgrade test failure:

```text
Terraform will perform the following actions:
        
          # module.aks.azurerm_kubernetes_cluster.main will be updated in-place
          ~ resource "azurerm_kubernetes_cluster" "main" {
                id                                  = "/subscriptions/f7a632a5-49db-4c5e-9828-cd62cb753971/resourceGroups/876f024bb7fd76f9-rg/providers/Microsoft.ContainerService/managedClusters/prefix-876f024bb7fd76f9-aks"
                name                                = "prefix-876f024bb7fd76f9-aks"
                tags                                = {}
                # (30 unchanged attributes hidden)
        
              ~ auto_scaler_profile {
                  ~ new_pod_scale_up_delay           = "0s" -> "10s"
                  ~ skip_nodes_with_local_storage    = false -> true
                    # (15 unchanged attributes hidden)
                }
        
                # (9 unchanged blocks hidden)
            }
```

If our users upgrade the module's version to the merged version then they would face such change. I would suggest adding two new variables with default value `10s` and `true`.

lonegunmanb (changes requested):
We need a new toggle variable for this new added dynamic block to avoid surprise change.

DavidSpek:
@lonegunmanb Sorry for the late reply, we were out of office during the holidays. Before I make the above change, would you otherwise prefer a single map variable with all the settings for the autoscaling profile, rather than a dedicated variable for each autoscaler profile setting?

DavidSpek:
@zioproto Maybe you can give your opinion about having a single map variable to configure the autoscaling profile rather than a separate variable for each field.

lonegunmanb:
> @lonegunmanb Sorry for the late reply, we were out of office during the holidays. Before I make the above change, would you otherwise prefer a single map variable with all the settings for the autoscaling profile, rather than a dedicated variable for each autoscaler profile setting?

Good point! I'm doing such practices in new created module like [this one](https://github.com/lonegunmanb/terraform-azurerm-virtual-machine/blob/main/variables.tf#L29-L61) by using `object` type with [optional attributes](https://developer.hashicorp.com/terraform/language/expressions/type-constraints#optional-object-type-attributes). The new `object` type solves some `map`'s issues so I would prefer `object`.

But for aks module, my concern is we need keep consistency among our variables. Maybe someday we could publish a major release, converting variables into `object` types, but for now I would say let's preserve the status quo.

test/unit/unit_test.go
```diff
@@ -199,6 +199,33 @@ func TestInvalidVarsForAutomaticUpgrades(t *testing.T) {
 	}
 }
 
+func TestScaleDownDelayAfterDeleteNotSetShouldUseScanInterval(t *testing.T) {
+	vars := dummyRequiredVariables()
+	test_helper.RunE2ETest(t, "../../", "unit-test-fixture", terraform.Options{
+		Upgrade: false,
+		Vars:    vars,
```
lonegunmanb, line 206: Since we didn't modify `vars`, I think maybe we can inline `vars` here to:

```go
Vars: dummyRequiredVariables(),
```

DavidSpek:
@lonegunmanb Should be all good now.

lonegunmanb (approved):
Thanks @DavidSpek, LGMT! 🚀

PR #279, opened 2022-12-02
Upgrade `terraform-module-test-helper` lib so we can get rid of override file to execute version upgrade test

lonegunmanb:
## Describe your changes

For now we need to put an `override.tf.tplt` file for every example folder to execute version upgrade test, because we use Terraform override file to modify module's source. Now we've upgrade `terraform-module-test-helper` lib, it can modify the hcl file directly to redirect the module's source.

## Issue number

#000

Thanks for your cooperation!


jiaweitao001 (approved):
LGTM. Thanks.

PR #281, opened 2022-12-09
feat: add var automatic_channel_upgrade

### LINKED ISSUE #280: Support for Auto-upgrade feature

### Description

Hello,

I would like to have auto upgrade cluster feature

Thanks

azurerm_kubernetes_cluster

```hcl
module "aks" {
...
automatic_channel_upgrade  = "none | patch | rapid | node-image | stable"
...
}
```

### References

https://learn.microsoft.com/en-us/azure/aks/auto-upgrade-cluster

the-technat:
Since AKS now supports automatic upgrades, we @swisspost like to use this feature. I addressed the concerns described in the issue with the following tests:

- Created cluster on `v1.23.8` using `examples/startup`.
- Configured `automatic_channel_upgrade=patch` using Terraform
- Manually upgrade cluster to `v1.23.12` using Azure CLI
- Rerun Terraform

=> As expected Terraform will downgrade the cluster to `v1.23.8`, same must the case when automatically patching minor versions and specifying the version.

So the correct usage would be that one specifies only the minor version (e.g `1.23`) and enables automatic patch upgrade or omits the version field completely and enables automatic minor upgrades. But since Terraform variable validation can't reference other variables I can't catch misconfiguration there. But If I'm honest, it makes absolutely sense to me, that you don't specify the patch version if you want automatic upgrades.

Note: the default value must be `null` and note `none` as the Provider docs would suggest. I got errors in validation when trying to pass `none` as default value.

## Issue number

Closes #280 


the-technat:
> Potential Breaking Changes in [07aef57](https://github.com/Azure/terraform-azurerm-aks/commit/07aef578a2cd035ab37576c9481ad2b14f19d6ce): [update] "Variables.log_analytics_solution_id.Nullable" from 'true' to '' [update] "Variables.log_analytics_workspace.Nullable" from 'true' to '' [update] "Variables.log_analytics_workspace_resource_group_name.Nullable" from 'true' to ''

This was changed due to tflint complaining that nullable is by default set to true.

variables.tf
```diff
@@ -93,6 +93,20 @@ variable "api_server_authorized_ip_ranges" {
   default     = null
 }
 
+variable "automatic_channel_upgrade" {
+  type        = string
+  default     = null
+  description = "(Optional) The upgrade channel for this Kubernetes Cluster. Possible values are  `patch`, `rapid`, `node-image` and `stable`. By default automatic-upgrades are turned off. See [the documentation](https://learn.microsoft.com/en-us/azure/aks/auto-upgrade-cluste) for more information"
+  validation {
```
zioproto, line 100: You should use the function `contains` here:

```
validation {
    condition = (
      var.automatic_channel_upgrade == null || (
        contains(["patch", "stable", "rapid", "node-image"], var.automatic_channel_upgrade)
      )
    )
    error_message = "`automatic_channel_upgrade`'s possible values are `patch`, `stable`, `rapid` or `node-image`."
  }
```

variables.tf
```diff
@@ -232,7 +246,6 @@ variable "log_analytics_solution_id" {
   type        = string
   description = "(Optional) Existing azurerm_log_analytics_solution ID. Providing ID disables creation of azurerm_log_analytics_solution."
   default     = null
-  nullable    = true
```
zioproto, line 235: I understand you removed this line because of a `tflint` error. But we should fix the CI in this case. Can you revert these 3 changes where you remove the lines `nullable = true` from the variables that are not in scope with this PR ? thanks

README.md
```diff
@@ -284,6 +284,7 @@ No modules.
...
+| <a name="input_automatic_channel_upgrade"></a> [automatic\_channel\_upgrade](#input\_automatic\_channel\_upgrade)                                                             | (Optional) The upgrade channel for this Kubernetes Cluster. Possible values are  `patch`, `rapid`, `node-image` and `stable`. By default automatic-upgrades are turned off. See [the documentation](https://learn.microsoft.com/en-us/azure/aks/auto-upgrade-cluste) for more information                                        | `string`                                                                                                                                                                                                                   | `null`                      |    no    |
```
zioproto, line 287: The link to the documentation is wrong: you are missing a `r`
https://learn.microsoft.com/en-us/azure/aks/auto-upgrade-cluste

zioproto: Could you also document that using this variable requires that you cannot specify a patch version in `kubernetes_version` ?

variables.tf
zioproto: Commenting on the tests have done earlier, would it work to validate like this ?
```
validation {
  condition = (
    var.automatic_channel_upgrade == null || (
      contains(["patch", "stable", "rapid", "node-image"], var.automatic_channel_upgrade) &&
      matches("^(?:[0-9]\\.){2}[0-9]$", var.kubernetes_version) == false
    )
  )
  error_message = "`automatic_channel_upgrade`'s possible values are `patch`, `stable`, `rapid` or `node-image` and `kubernetes_version` must not include a patch version number."
}
```

zioproto: Should the node-image channel be independent from the Kubernetes patch version check ?

zioproto (changes requested):

variables.tf
zioproto: Hey sorry I totally missed that you cannot validate a variable referencing another variable value. But you could move this logic to the `azurerm_kubernetes_cluster` resource to validate at runtime.

the-technat:
Hey @zioproto 

Thanks for your review. I'll take a look at the requested changes. Might get tomorrow until I've time.

README.md
the-technat: Sure, added a sentence about this and fixed the typo.

variables.tf
the-technat: @zioproto I've used `contains` in the variable validation (to get fast feedback if the value is valid) and added a second more extensive precondition to the `azurerm_kubernetes_cluster` resource. Could you have a look at it?

main.tf
```diff
@@ -220,6 +221,21 @@ resource "azurerm_kubernetes_cluster" "main" {
...
+        contains(["patch"], var.automatic_channel_upgrade) &&
+        matches("^[0-9]{1,}\\.[0-9]{1,}$", var.kubernetes_version)
+        ) || (
+        // or we have set it to rapid/stable and don't specify the version at all
+        contains(["rapid", "stable", "node-image"], var.automatic_channel_upgrade) &&
+        var.kubernetes_version == null
+        )
+      )
+      error_message = "Either disable automatic upgrades, only specify up to the minor version when using `automatic_channel_upgrade=patch` or don't specify `kubernetes_version` at all when using `automatic_channel_upgrade=stable|rapid|node-image`."
+
```
lonegunmanb, line 237: Would you please remove this blank line?

variables.tf
```diff
@@ -93,6 +93,20 @@ variable "api_server_authorized_ip_ranges" {
   default     = null
 }
 
+variable "automatic_channel_upgrade" {
+  type        = string
+  default     = null
+  description = "(Optional) The upgrade channel for this Kubernetes Cluster. Possible values are  `patch`, `rapid`, `node-image` and `stable`. By default automatic-upgrades are turned off. Note that you cannot use the `patch` upgrade channel and still specify the patch version using `kubernetes_version`. See [the documentation](https://learn.microsoft.com/en-us/azure/aks/auto-upgrade-cluster) for more information"
```
lonegunmanb, line 99: I think we've got an extra space here:

```text
Possible values are  `patch`
```

Should be:

```text
Possible values are `patch`
```

variables.tf
```diff
@@ -93,6 +93,20 @@ variable "api_server_authorized_ip_ranges" {
...
+  validation {
+    condition = (
+      var.automatic_channel_upgrade == null || (
```
lonegunmanb, line 102: Since the [Terraform's binary operator is not short-circuit](https://github.com/hashicorp/terraform/issues/24128), we would have an issue when `var.automatic_channel_upgrade` is `null`, though `var.automatic_channel_upgrade == null` is `true`, Terraform would continue to evaluate the value of `contains(["patch", "stable", "rapid", "node-image"], var.automatic_channel_upgrade)`, which equals to `contains(["patch", "stable", "rapid", "node-image"], null)`, and that would caused an error, the `contains` function dosen't accept `null` as parameter.

I would recommend the following expression:

```hcl
var.automatic_channel_upgrade == null ? true : (contains(["patch", "stable", "rapid", "node-image"], var.automatic_channel_upgrade))
```

main.tf
```diff
@@ -220,6 +221,21 @@ resource "azurerm_kubernetes_cluster" "main" {
       condition     = !(var.microsoft_defender_enabled && !var.log_analytics_workspace_enabled)
       error_message = "Enabling Microsoft Defender requires that `log_analytics_workspace_enabled` be set to true."
     }
+    precondition {
+      // either we have automatic upgrades disabled
+      condition = (var.automatic_channel_upgrade == null || (
```
lonegunmanb, line 226: The same [short-circuit](https://github.com/hashicorp/terraform/issues/24128) issue we've got in `variable`'s `validation` block.

main.tf
```diff
@@ -220,6 +221,21 @@ resource "azurerm_kubernetes_cluster" "main" {
       condition     = !(var.microsoft_defender_enabled && !var.log_analytics_workspace_enabled)
       error_message = "Enabling Microsoft Defender requires that `log_analytics_workspace_enabled` be set to true."
     }
+    precondition {
```
lonegunmanb, line 224: Since the `condition` for this `precondition` is complex, I would rather to create a new `local` ,in `locals.tf` file, and add some unit tests for this condition check. Would you please refer the current unit test inside `test/unit` folder and help us by adding three unit test cases to cover this `precondition`? Thanks!

main.tf
```diff
@@ -220,6 +221,21 @@ resource "azurerm_kubernetes_cluster" "main" {
...
+      // either we have automatic upgrades disabled
+      condition = (var.automatic_channel_upgrade == null || (
+        // or we have set it to patch and only specify the minor version
+        contains(["patch"], var.automatic_channel_upgrade) &&
+        matches("^[0-9]{1,}\\.[0-9]{1,}$", var.kubernetes_version)
```
lonegunmanb, line 229: Is there any special reason we should use this regex here instead of determining whether `var.kubernetes_version` is `null` or empty?

lonegunmanb (changes requested):
Hi @the-technat thanks for opening this pr! The `pr-check` job failed so would you please run the `pre-commit` and `pr-check` job on your machine as the [document](https://github.com/Azure/terraform-azurerm-aks#pre-commit--pr-check--test) instructed then try again?

I also recommend adding this new variable in `examples/startup`'s code so our e2e test could cover it.  Again, thanks for your contribution!

zioproto:
@the-technat I have to go back on the `nullable = true` removals in the `variables.tf` file. Those changes were correct. The CI was updated, and now `tflint` shows as redundant to have `nullable = true` when the default value is already null.

Please dont fix it in this PR. Could you please rebase your branch on the current `master` branch ? This will solve that issue.

variables.tf
the-technat: Good point, I've read about this and then forgot it again. Switched to your snippet.

main.tf
the-technat: Good point. I'll move it into `locals.tf` and add a unit test.

variables.tf
the-technat: Resolving this for now since it's outdated with the changes suggested by @lonegunmanb.

main.tf
the-technat: Exactly. Simply checking if it's null is only viable if someone set `automatic_channel_upgrade=stable|rapid` since then AKS updates to new minor version too and noone only specifies the major version I guess. But with automatic patch upgrades we want to make sure that a valid minor version (e.g `1.24` or `1.25`) is specified.

the-technat:
@zioproto, @lonegunmanb I addressed some of your suggestions already, rebased the PR and fixed the pre-commit/pr-test check. 

I'm happy to add some unit tests as well but first we need to find the correct way of validation. @lonegunmanb suggested moving the checks into `locals.tf` cause they are quite complex. I like this idea but locals can only control the value of something, not if it get's applied or not, so what about doing the following:

- remove `automatic_channel_upgrade` variable completely 
- set the `automatic_channel_upgrade` filed on the cluster resource based on the version the user specifies
  - if `kubernetes_version=1.23.12`, set `automatic_channel_upgrade=null`
  - if `kubernetes_version=1.23`, set `automatic_channel_upgrade=patch`
  - if `kubernetes_version=null`, set `automatic_channel_upgrade=stable` (or rapid, maybe controlled via extra input var)

main.tf
lonegunmanb: >I though about your suggestion of putting the check into locals.tf but can locals contain a precondition check?

Could we extract the expression into a local varialbe with `bool` type? Like:

```hcl
locals{
automatic_channel_upgrade_check = (var.automatic_channel_upgrade == null || (
contains(["patch"], var.automatic_channel_upgrade) &&
        regex("^[0-9]{1,}\\.[0-9]{1,}$", var.kubernetes_version)
        ) || (
        # or we have set it to rapid/stable and don't specify the version at all
        contains(["rapid", "stable", "node-image"], var.automatic_channel_upgrade) &&
        var.kubernetes_version == null
        )
      )
}
```

Then the `precondition` could be:

```hcl
precondition {
      // either we have automatic upgrades disabled
      condition = local.automatic_channel_upgrade_check
...
```

main.tf
lonegunmanb: Since I'm not an aks expert, I'll leave the shoot to @zioproto , you make the call.

lonegunmanb:
> @zioproto, @lonegunmanb I addressed some of your suggestions already, rebased the PR and fixed the pre-commit/pr-test check.
> 
> I'm happy to add some unit tests as well but first we need to find the correct way of validation. @lonegunmanb suggested moving the checks into `locals.tf` cause they are quite complex. I like this idea but locals can only control the value of something, not if it get's applied or not, so what about doing the following:
> 
> * remove `automatic_channel_upgrade` variable completely
> * set the `automatic_channel_upgrade` filed on the cluster resource based on the version the user specifies
>   
>   * if `kubernetes_version=1.23.12`, set `automatic_channel_upgrade=null`
>   * if `kubernetes_version=1.23`, set `automatic_channel_upgrade=patch`
>   * if `kubernetes_version=null`, set `automatic_channel_upgrade=stable` (or rapid, maybe controlled via extra input var)

I think we could just extract the expression we're using now for the `condition` in the `precondition` block into a new local variable, and this expression would only referred variable, so we can mock different values to test different cases. Could we do that?

the-technat:
@lonegunmanb ah now I see what you meant. Yeah that should work. I'll try to do it.

main.tf
the-technat: Should be solved now.

main.tf
the-technat: I've moved the condition into a bool var.

the-technat:
@lonegunmanb I added a table-driven unit test to ensure that the upgrade check produces the right result. Is this what you expected?

the-technat:
Tests still fail but locally they pass.

lonegunmanb:
> Tests still fail but locally they pass.

No worries, I'll check the test.

lonegunmanb:
> Tests still fail but locally they pass.

Apology for the inconvenience @the-technat , after a quick glance your unit tests are good, the failure was caused by Checkov check. I'll keep digging.

lonegunmanb:
Hi @the-technat , thanks for your patience, it looks like Checkov (the policy checking tool we're using now in this pipeline) was broken, to fix the Checkov we need upgrade Checkov's version in our CI runner image, but that brought another issue, in newer Checkov they've introduced new policies for Aks resource, which this module failed to comply for now. I don't want to make this pr too complex, so I want to skip these new policies, would you please help me by adding this [config file](https://gist.github.com/lonegunmanb/1f20b4d49129f20f2ba580d8da87725e) with you pr so the Checkov would just skip these policies? The new checkov file name should be `.checkov_config.yaml`, that would fix our issue.

the-technat:
@lonegunmanb no problem, we all well know flaky CI tests 😉. I've added the config file as you recommended and now the tests seem to pass here as well.

locals.tf
```diff
@@ -1,4 +1,15 @@
 locals {
+  # automatic upgrades are either:
+  # - null 
+  # - patch, but then the kubernetes_version musnt't specify a patch number
+  # - rapid/stable/node-image, but then the kubernetes_version must be null
+  automatic_channel_upgrade_check = var.automatic_channel_upgrade == null ? true : (
+    contains(["patch"], var.automatic_channel_upgrade) &&
+    can(regex("^[0-9]{1,}\\.[0-9]{1,}$", var.kubernetes_version))
+    ) == true ? true : (
```
lonegunmanb, line 9: Could we omit `== true` here?

```text
) ? true : (
```

lonegunmanb (changes requested):
Thanks @the-technat for your update, almost LGTM, once the e2e test passed, only a tiny change request to solve, one extra commit contains only one line change would make this pr been merged.

locals.tf
the-technat: Sure, I've changed it.

lonegunmanb:
> Based on image
> 
> I switched the version to `1.24` in e2e tests. @lonegunmanb would you rerun them?

With pleasure!

lonegunmanb (approved):
Thanks @the-technat , LGTM! 🚀

PR #283, opened 2022-12-15
auto-upgrade: variable orchestrator_version to null

zioproto:
## Describe your changes

This is an addition to PR #281

In AKS is possible to upgrade the control plane and the kubernetes nodes independently. The Terraform resource `azurerm_kubernetes_cluster` exposes this with 2 variables `kubernetes_version` and  `orchestrator_version`.

In PR #281 the lifecycle validation check takes into account only `kubernetes_version`.

## Issue number

Related to  #280

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -1,15 +1,13 @@
 locals {
   # automatic upgrades are either:
-  # - null 
-  # - patch, but then the kubernetes_version musnt't specify a patch number
-  # - rapid/stable/node-image, but then the kubernetes_version must be null
-  automatic_channel_upgrade_check = var.automatic_channel_upgrade == null ? true : (
-    contains(["patch"], var.automatic_channel_upgrade) &&
-    can(regex("^[0-9]{1,}\\.[0-9]{1,}$", var.kubernetes_version))
-    ) ? true : (
-    contains(["rapid", "stable", "node-image"], var.automatic_channel_upgrade) &&
-    var.kubernetes_version == null
-  )
+  # - null
+  # - patch, but then the kubernetes_version must not specify a patch number and orchestrator_version must be null
+  # - rapid/stable/node-image, but then the kubernetes_version and the orchestrator_version must be null
+  automatic_channel_upgrade_check = var.automatic_channel_upgrade == null ? true : var.orchestrator_version == null && (
+    (contains(["patch"], var.automatic_channel_upgrade) && can(regex("^[0-9]{1,}\\.[0-9]{1,}$", var.kubernetes_version)))
+    || (contains(["rapid", "stable", "node-image"], var.automatic_channel_upgrade) && var.kubernetes_version == null
+  ))
+
   # Abstract the decision whether to create an Analytics Workspace or not.
   create_analytics_solution  = var.log_analytics_workspace_enabled && var.log_analytics_solution_id == null
   create_analytics_workspace = var.log_analytics_workspace_enabled && var.log_analytics_workspace == null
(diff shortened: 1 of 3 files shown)


the-technat (approved):
Thanks for coming up with this! I totally forgot about the `orchestrator_version`, but now that we told about the actually behavior of the nodes, it makes sense to omit this when automatic upgrades are turned on.

locals.tf
```diff
@@ -1,14 +1,16 @@
...
     contains(["patch"], var.automatic_channel_upgrade) &&
-    can(regex("^[0-9]{1,}\\.[0-9]{1,}$", var.kubernetes_version))
+    can(regex("^[0-9]{1,}\\.[0-9]{1,}$", var.kubernetes_version)) &&
+    var.orchestrator_version == null
     ) ? true : (
     contains(["rapid", "stable", "node-image"], var.automatic_channel_upgrade) &&
-    var.kubernetes_version == null
+    var.kubernetes_version == null &&
+    var.orchestrator_version == null
   )
```
lonegunmanb, line 14: I would like to try:

```hcl
automatic_channel_upgrade_check = var.automatic_channel_upgrade == null ? true : var.orchestrator_version == null && (
    (contains(["patch"], var.automatic_channel_upgrade) && can(regex("^[0-9]{1,}\\.[0-9]{1,}$", var.kubernetes_version)))
    || (contains(["rapid", "stable", "node-image"], var.automatic_channel_upgrade) && var.kubernetes_version == null
  ))
```

main.tf
```diff
@@ -241,7 +241,7 @@ resource "azurerm_kubernetes_cluster" "main" {
     }
     precondition {
       condition     = local.automatic_channel_upgrade_check
-      error_message = "Either disable automatic upgrades, only specify up to the minor version when using `automatic_channel_upgrade=patch` or don't specify `kubernetes_version` at all when using `automatic_channel_upgrade=stable|rapid|node-image`."
+      error_message = "Either disable automatic upgrades, or only specify up to the minor version when using `automatic_channel_upgrade=patch` or don't specify `kubernetes_version` at all when using `automatic_channel_upgrade=stable|rapid|node-image`. With automatic upgrades orchestrator_version must be set to null."
```
lonegunmanb, line 244: Could we back quote the variable's name inside the message here?:

```hcl
error_message = "Either disable automatic upgrades, or only specify up to the minor version when using `automatic_channel_upgrade=patch` or don't specify `kubernetes_version` at all when using `automatic_channel_upgrade=stable|rapid|node-image`. With automatic upgrades `orchestrator_version` must be set to `null`."
```

lonegunmanb (changes requested):
Thanks @zioproto for opening this pr and thanks @the-technat for participating our discussion. Almost LGTM but I'd like to simplify the logic expression. And since this expression is very complex, would you please add some unit test cases to cover this expression? I would recommend adding the new unit tests first to prove your expression works, then try my suggestion with protection from these tests. With the protection from your test cases, we could change or refactor this expression in the future. Thanks @zioproto !

main.tf
zioproto: done

locals.tf
zioproto: Done ! Implemented new unit tests to test both when it should return automatic_channel_upgrade_check `true` and automatic_channel_upgrade_check `false`

zioproto:
@lonegunmanb please take a look again

lonegunmanb:
Hello @zioproto , I'm sorry we've met an issue on our pipeline so we have to upgrade the Github action yaml files in this repo, would you please merge your branch with the latest master branch so we can run the e2e test? Apology for the inconvenience, we're still refactoring the pipeline.

zioproto:
@lonegunmanb I rebased my commit on top of the current master branch. Thanks

PR #288, opened 2023-01-06
feat: Implement support for KMS arguments

mkilchhofer:
## Describe your changes

This implements [Add Key Management Service (KMS) etcd encryption to an Azure Kubernetes Service (AKS) cluster](https://learn.microsoft.com/en-us/azure/aks/use-kms-etcd-encryption).

Feature is not yet available in azurerm provider, but also opened a PR over there:
- https://github.com/hashicorp/terraform-provider-azurerm/pull/19893

## Issue number

`-`

Thanks for your cooperation!


main.tf
```diff
@@ -136,6 +136,16 @@ resource "azurerm_kubernetes_cluster" "main" {
       subnet_id    = var.ingress_application_gateway_subnet_id
     }
   }
+  dynamic "key_vault_kms" {
+    for_each = var.key_vault_kms_enabled ? ["key_vault_kms"] : []
```
lonegunmanb, line 140: Since we could turn `enabled` to `false` meanwhile keeping this `key_vault_kms`, I would recommend the following toggle expression:

```hcl
for_each = var.key_vault_kms_enabled != null ? ["key_vault_kms"] : []
```

lonegunmanb (changes requested):
Thanks @mkilchhofer for opening this pr and submitting a new resource pr to terraform-azurerm-provider's repo, I've left some comments on provider's pr, once the pr has been merged into the provider, we can continue the code review for this pr.


Btw the review comments for your provider's pr are my personal suggestions, there's no guarantee that your pr would be accepted by HashiCorp if you accepted my suggestion.

main.tf
mkilchhofer: refactored this anyways (source was the upstream implemantation)

mkilchhofer:
PR is now updated after the upstream provider PR got merged AND released ;-)

@lonegunmanb can you please review again?

variables.tf
```diff
@@ -335,6 +335,24 @@ variable "key_vault_secrets_provider_enabled" {
   nullable    = false
 }
 
+variable "kms_enabled" {
+  type        = bool
+  description = "(Optional) Enable Azure KeyVault Key Management Service."
+  default     = false
```
lonegunmanb, line 341: Would an extra `nullable = false` here be better?

variables.tf
```diff
@@ -335,6 +335,24 @@ variable "key_vault_secrets_provider_enabled" {
...
+variable "kms_key_vault_key_id" {
+  type        = string
+  description = "(Optional) Identifier of Azure Key Vault key. When Azure Key Vault key management service is enabled, this field is required and must be a valid key identifier."
+  default     = null
+}
+
+variable "kms_key_vault_network_access" {
+  type        = string
+  description = "(Optional) Network Access of Azure Key Vault. Possible values are: `Private` and `Public`. If not set, defaults to type `Public`."
+  default     = null
```
lonegunmanb, line 353: Should the default value be `Public` here?

lonegunmanb (changes requested):
Thanks @mkilchhofer for updating this pr, some issues need to be solved.

It would be nice if we could have an new example to demonstrate this new feature. We have already created a KeyValut in `startup` example, but for disk encryption key management for now, would you please estimate whether if we can make that KeyVault work for this new feature? We might need to update `azurerm` provider's restriction in `providers.tf` file in `startup` folder. Or we can just crate a new example folder to do so, your call.

Again, thanks for your contribution!

variables.tf
mkilchhofer: Not sure if it is good practice to bubble defaults from the provider up to the module.
It's already bubbled up from go-sdk to terraform provider. So if it's set to `null`, it defaults to `Public` on the provider level.

But I am also fine do default it also here. :-)

variables.tf
mkilchhofer: I can do this, sure. 👍

main.tf
```diff
@@ -159,6 +159,14 @@ resource "azurerm_kubernetes_cluster" "main" {
       subnet_id    = var.ingress_application_gateway_subnet_id
     }
   }
+  dynamic "key_management_service" {
+    for_each = var.kms_enabled ? ["key_management_service"] : []
```
zioproto, line 163: KMS etcd encryption doesn't work with system-assigned managed identity. You should check for `var.kms_enabled` and for `var.identity_type` to be `UserAssigned`.

zioproto (changes requested):

variables.tf
mkilchhofer: Done

main.tf
mkilchhofer: Done

mkilchhofer:
I cannot test the examples as we @swisspost have no subscriptions which allows usage of public IPs.
Can someone test that for us?

zioproto:
@mkilchhofer I can do that ! thanks

zioproto:
@mkilchhofer  could you please run the `pre-commit` step on your local machine ?:

```
docker run --rm -v $(pwd):/src -w /src mcr.microsoft.com/azterraform:latest make pre-commit
```

You patched the example and you have duplicated definitions for `identity_type` and `identity_ids`. See the error:

```
Error: Attribute redefined
│
│   on examples/named_cluster/main.tf line 88, in module "aks_cluster_name":
│   88:   identity_type         = "UserAssigned"
│
│ The argument "identity_type" was already set at
│ examples/named_cluster/main.tf:66,3-16. Each argument may be set only once.
╵

╷
│ Error: Attribute redefined
│
│   on examples/named_cluster/main.tf line 89, in module "aks_cluster_name":
│   89:   identity_ids          = [azurerm_user_assigned_identity.test.id]
│
│ The argument "identity_ids" was already set at
│ examples/named_cluster/main.tf:65,3-15. Each argument may be set only once.
```

examples/named_cluster/main.tf
```diff
@@ -77,4 +77,9 @@ module "aks_cluster_name" {
   rbac_aad                          = true
   rbac_aad_managed                  = true
   role_based_access_control_enabled = true
+
+  # KMS encrption
+  key_vault_kms_enabled = true
```
zioproto, line 82: These variable name is invalid. Should be just `kms_enabled`

examples/named_cluster/main.tf
```diff
@@ -77,4 +77,9 @@ module "aks_cluster_name" {
...
+  key_vault_kms_key_id  = azurerm_key_vault_key.kms.id
```
zioproto, line 83: This should be `kms_key_vault_key_id` according to your change in `variable.tf`

zioproto (changes requested):
Could you patch only the `named_cluster` example ? 

We dont need to the KMS in the others, it would also require more work because the needed `UserAssigned` identity_type is not set in the others. Thanks

examples/named_cluster/main.tf
mkilchhofer: Yea. Sorry, copy pasta error before refactoring :D

examples/named_cluster/main.tf
mkilchhofer: Also copy pasta error before refacotoring :/

zioproto:
commit 70121785da3e7b0a78393f44b0c1a9c87566902b is now passing both `pre-commit` and `pr-check` steps.

@lonegunmanb could you please review and run the e2e tests ? thanks

.checkov_config.yaml
```diff
@@ -11,6 +11,7 @@ quiet: true
 secrets-scan-file-type: []
 skip-check:
   - CKV_GHA_3
+  - CKV_AZURE_112
```
zioproto, line 14: I believe this is OK because not all customers require the use of HSM for Azure Key Vault

examples/named_cluster/kms.tf
```diff
@@ -0,0 +1,29 @@
...
+  key_size        = 2048
+
+  depends_on = [
+    azurerm_key_vault_access_policy.current_user
+  ]
+}
+
+resource "azurerm_key_vault_access_policy" "kms" {
+  key_vault_id = azurerm_key_vault.des_vault.id
+  object_id    = azurerm_user_assigned_identity.test.id
```
lonegunmanb, line 23: We've got an e2e test error:

```text
TestExamplesNamedCluster 2023-01-27T04:53:34Z logger.go:66: │ Error: expected "object_id" to be a valid UUID, got /subscriptions/f7a632a5-49db-4c5e-9828-cd62cb753971/resourceGroups/8df3563089d5cbb3-rg/providers/Microsoft.ManagedIdentity/userAssignedIdentities/8df3563089d5cbb3-identity
TestExamplesNamedCluster 2023-01-27T04:53:34Z logger.go:66: │ 
TestExamplesNamedCluster 2023-01-27T04:53:34Z logger.go:66: │   with azurerm_key_vault_access_policy.kms,
TestExamplesNamedCluster 2023-01-27T04:53:34Z logger.go:66: │   on kms.tf line 23, in resource "azurerm_key_vault_access_policy" "kms":
TestExamplesNamedCluster 2023-01-27T04:53:34Z logger.go:66: │   23:   object_id    = azurerm_user_assigned_identity.test.id
```

Could you please try `object_id = azurerm_user_assigned_identity.test.principal_id`, then try to pass `TestExamplesNamedCluster` test inside `test/e2e` folder? Thanks!

lonegunmanb (changes requested):

variables.tf
mkilchhofer: 👍 will do

examples/named_cluster/kms.tf
mkilchhofer: Cannot run E2E here as I have no access to a unrestricted Azure Subscription.

But jup, `principal_id` will fix it halfway, we also need a `depends_on` inside the call of the module to ensure the managed identity has the correct access policy applied before we call the managedCluster API.

mkilchhofer: done

mkilchhofer:
@lonegunmanb can you trigger E2E again? 😇

lonegunmanb (changes requested):
Thanks @mkilchhofer for this update, we've met some issues when we ran the test.

We've added `network_acl` on the KeyVault to restrict the source IP which can access this Key Vault, but the Aks's public IP is unpredictable so the original code would meet a 403 error.

We need config the aks to access the KeyVault via private network as [this document](https://learn.microsoft.com/en-us/azure/aks/use-kms-etcd-encryption#create-an-aks-cluster-with-private-key-vault-and-enable-kms-etcd-encryption) described.

examples/named_cluster/main.tf
```diff
@@ -77,4 +77,12 @@ module "aks_cluster_name" {
   rbac_aad                          = true
   rbac_aad_managed                  = true
   role_based_access_control_enabled = true
+
```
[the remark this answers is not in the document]
mkilchhofer: Service endpoint is not needed (at least in our scenario). I think its only needed if some part inside the VNET wants to access the keyvault.

examples/named_cluster/main.tf
```diff
@@ -77,4 +77,12 @@ module "aks_cluster_name" {
...
+  # KMS etcd encryption
+  kms_enabled          = true
+  kms_key_vault_key_id = azurerm_key_vault_key.kms.id
+
```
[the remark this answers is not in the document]
mkilchhofer: We at @swisspost use private connection. Should the examples also use the private one?
Thought that `bypass         = "AzureServices"` will allow AKS to communicate with vault but it seems that its not covered in that.

lonegunmanb (approved):
Thanks @mkilchhofer, LGTM! 🚀

mkilchhofer:
Thanks alot for your patience with me 👍

Issue #299, opened 2023-02-09
Support for [taints for default node group]


fabiomarinetti:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Up to the current version of the module there's no way of configuring taints on default node group.
azurerm providers >=3.38.0 give such support. Could you pls embed it?

### New or Affected Resource(s)/Data Source(s)

azurerm_kubernetes_cluster

### Potential Terraform Configuration

```hcl
...
agent_taints = [
  "key1=value1:NoSchedule",
  "key2=value2:NoSchedule"
]
...
```

Issue #85, opened 2021-01-06
Updating values forces create new resources 


sshah90:
Hello, This is a very easy module to work with but after deploying AKS cluster but when I try to apply some changes/updates it creates a new cluster. 

After debugging I figure out that there are some values that are by default set up as "null" which causing this re-creation.

Check Below screenshots from my plan

![Screen Shot 2021-01-06 at 1 01 43 PM](https://user-images.githubusercontent.com/26821376/103804298-ddc7d000-501f-11eb-881c-add1a06ae7ea.png)

As you can see `node_taints` (this var needs to be added in the module), `agents_availability_zones` should be `[]` but as its default value is null so if a user doesn't provide it so resource recreate aks.

If we don't set auto_scaling for the default node pool then max and min pod should be `0` not null and the same goes with `network_policy` as well.

Please update/add the following variable's default values.

- network_policy = "azure"
- agents_availability_zones = []
- agents_max_count = 0 (also change `0` [here](https://github.com/Azure/terraform-azurerm-aks/blob/master/main.tf#L36-L37))
- agents_min_count = 0 
- node_taints= [] (add new var in default_node_pool block)

yupwei68:
Hi @sshah90 , thanks for opening this issue. Would you provide your Azurerm version, terraform version, your hcl script and reproducing steps to help to reproduce this problem? Sorry that I can not reproduce this problem now.

yupwei68:
Hi @sshah90 ,sorry that I could not reproduce your error. 
I tested with the script below, and I `terraform apply`, and then `terraform plan`. It shows no changes.

```
module "aks" {
  source                         = "Azure/aks/azurerm"
  prefix                         = "prefix"
  resource_group_name            = azurerm_resource_group.test.name
  enable_log_analytics_workspace = false
  enable_kube_dashboard          = false
  net_profile_pod_cidr           = "10.1.0.0/16"
  depends_on                     = [azurerm_resource_group.test]
}
```

sshah90:
Hi @yupwei68, sorry for the late reply.

I have my code somewhere but not able to find it at this moment.

I will try to reproduce this issue over the weekend and let u know.

lonegunmanb:
I'm closing this issue due to no further response. Please feel free to reopen this issue if you can reproduce this issue with the latest version, thanks for your understanding @sshah90 .

PR #300, opened 2023-02-10
Add support for default node pool's `node_taints`

lonegunmanb:
## Describe your changes

This pr added support for `default_node_pool.node_taints` argument.

## Issue number

#299 

Thanks for your cooperation!


mloskot:
Shouldn't the default be `[]` instead of `null`? See https://github.com/Azure/terraform-azurerm-aks/issues/85

I'm observing the following plan changes (Terraform 1.3.7, hashicorp/azurerm v3.43.0):

```
Note: Objects have changed outside of Terraform
...
        name                    = "nplin1"
        node_count              = 2
      + node_taints             = []
        tags                    = {
      ...
```

lonegunmanb:
Thanks for asking @mloskot, for this argument I've checked the [schema definition](https://github.com/hashicorp/terraform-provider-azurerm/blob/main/internal/services/containers/kubernetes_nodepool.go#L166-L173) in AzureRM provider:

```go
"node_taints": {
						Type:     pluginsdk.TypeList,
						ForceNew: true,
						Optional: true,
						Elem: &pluginsdk.Schema{
							Type: pluginsdk.TypeString,
						},
					},
```

`Optional: true` means the default value for this argument is `null`. As for the change you've seen in the plan, you can export the tfplan as json and check whether if there's any action generated for that change, as I've done in our [test framework](https://github.com/Azure/terraform-module-test-helper/blob/main/upgradetest.go#L132-L160). If there's no action then we don't need to worry about it.

mloskot:
@lonegunmanb Thanks for the explanation and suggestions. I'll try it out.

jiaweitao001 (approved):
🚢

Issue #335, opened 2023-03-24
Setting automatic_channel_upgrade to "patch" leads to Terraform state drift


zioproto:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Greenfield/Brownfield provisioning

greenfield

### Terraform Version

1.4.2

### Module Version

6.7.1

### AzureRM Provider Version

v3.49.0

### Affected Resource(s)/Data Source(s)

azurerm_kubernetes_cluster

### Terraform Configuration Files

```hcl
module "aks" {
  source                           = "Azure/aks/azurerm"
  version                          = "6.7.1"
  kubernetes_version               = "1.25"
  orchestrator_version             = null
  automatic_channel_upgrade        = "patch"
[..CUT..]
}
```

### tfvars variables values

```hcl
none
```

### Debug Output/Panic Output

```shell
~ resource "azurerm_kubernetes_cluster" "main" {
id = "{REDACTED}"
~ kubernetes_version = "1.25.6" -> "1.25"
```

### Expected Behaviour

Terraform should not have state drift when an automatic upgrade happens in the `patch` channel.

### Actual Behaviour

When creating the cluster the AKS API returns this values:

```
 az aks show --name istio-aks -g istio-aks -o json | jq ".kubernetesVersion, .currentKubernetesVersion"
"1.25"
"1.25.5"
```

After an automatic upgrade ( `automatic_channel_upgrade=patch` ) the AKS API will return something like:

```
 az aks show --name istio-aks -g istio-aks -o json | jq ".kubernetesVersion, .currentKubernetesVersion"
"1.25.6"
"1.25.6"
```

Because Terraform saved in the state:

```
   "kubernetes_version": "1.25",
```

This will cause a Terraform state drift at the next `terraform apply` operation.

### Steps to Reproduce

`terraform apply` with this minimal config:

```
module "aks" {
  source                           = "Azure/aks/azurerm"
  version                          = "6.7.1"
  kubernetes_version               = "1.25"
  orchestrator_version             = null
  automatic_channel_upgrade        = "patch"
[..CUT..]
}
```

Wait for an automatic upgrade to happen and run again `terraform apply`

PR #336, opened 2023-03-27
[WIP] - Ignore changes on `kubernetes_version` from outside of Terraform

lonegunmanb:
Using azapi update resource to update aks kubernetes_version when use r changed it at Terraform side, ignore all other changes outside of Terraform

## Issue number

#335

Thanks for your cooperation!


main.tf
```diff
@@ -391,6 +393,27 @@ resource "azurerm_kubernetes_cluster" "main" {
   }
 }
 
+resource "null_resource" "kubernetes_version_keeper" {
+  triggers = {
+    version = var.kubernetes_version
+  }
+}
+
+resource "azapi_update_resource" "aks_cluster_post_create" {
```
the-technat, line 402: So an update of AKS would be done through this resource rather than the actual cluster resource?

viters:
Does setting specific version like `1.25.6` block the Azure from applying `automatic_channel_upgrade=patch`?

If yes - IMHO it is a problem on the Azure side and I agree with your solution, which fixes the Azure issue. From Terraform's perspective it is a clever workaround, to be frank I am surprised that Terraform does not have solution for such situations.

I would expect that setting specific version does not block the automatic upgrades - and therefore I would say "_if Azure upgrades your cluster and Terraform informs you that the state has changed, **just adjust your Terraform files and state to match up**_". We should try to resemble the state of infrastructure as close as we can with Terraform, rather than covering it up.

zioproto:
@viters I opened https://github.com/Azure/AKS/issues/3573 to give feedback for the AKS API. Please feel free to add feedback for the Azure AKS team there. Thanks !

viters:
@zioproto Thanks for link, but I feel like we are talking about different problems. Have you maybe tested: if you create a cluster with version `1.25.5` and set `automatic_channel_upgrade=patch`, will it upgrade to `1.25.6`?

the-technat:
> @zioproto Thanks for link, but I think we are tackling different problems. Have you maybe tested: if you create a cluster with version `1.25.5` and set `automatic_channel_upgrade=patch`, will it upgrade to `1.25.6`? If it will, then I think AKS works properly (apart from API inconsistency that you mentioned) and I think this PR's change is unwanted and #335 should be marked as won't fix.

I tested this behavior some months ago, it will detect a state drift and try to downgrade the cluster.

matt-FFFFFF:
Good workaround @lonegunmanb

viters:
@the-technat 
Ok, so AKS works properly. Now, you run a plan for your AKS and you see that the version has been updated. You have two options: downgrade the cluster to meet your Terraform requirements or update your Terraform requirements to resemble the infrastructure. I see your point and room for error, but I just think that Terraform is here document the infrastructure and you should verify what plan does and adjust your files if needed.

zioproto:
> @zioproto Thanks for link, but I think we are tackling different problems. Have you maybe tested: if you create a cluster with version `1.25.5` and set `automatic_channel_upgrade=patch`, will it upgrade to `1.25.6`? If it will, then I think AKS works properly (apart from API inconsistency that you mentioned) and I think this PR's change is unwanted and #335 should be marked as won't fix.

In an ideal situation we should track in the Terraform state only  the `major` and `minor` version in the Terraform state and let AKS handle the changes in the `patch` version when the `automatic_channel_upgrade=patch`. 

Because of the behaviour documented in https://github.com/Azure/AKS/issues/3573 it was missed that the patch version is actually returned by the API and there is this collision with the information saved in the Terraform state.

viters:
> In an ideal situation we should track in the Terraform state only the major and minor version in the Terraform state and let AKS handle the changes in the patch version when the automatic_channel_upgrade=patch.

Right, now I see and I agree. Your point is that when you set `automatic_channel_upgrade=patch`, then you care only about major and minor. I missed it before.

The only concern I have left is, as this workaround is a breaking change, assuming that Azure would fix API in the future, reverting this workaround later will be a breaking change as well. Although I think there is not much to be done other than choosing to wait for Azure API adjustment, which may take some time to land.

zioproto:
> Although, I guess it is better to have workaround quickly rather than no-ETA waiting for Azure to adjust the API.

The ETA for an eventual AKS API change depends on the quality of feedback that we are able to write as a community on https://github.com/Azure/AKS/issues/3573

main.tf
lonegunmanb: @the-technat yes, since in this approach we rely on `null_resource`'s replacement to trigger the update, we cannot execute the update via azurerm resource since the whole resource would be re-created, so it must be a virtual resource like `azapi_update_resource`.

lonegunmanb:
As `automatic_channel_upgrade=patch` would trigger changes on patch version from the service side, this approach would ignore all these kind of patch upgrades, but once we've changed `kubernetes_version` on Terraform side, an update would be executed.

On the other hand, this pr should be considered as an unavoidable change, so the community's feedback is critical to us.

the-technat:
@viters just to give some feedback to your questions. I think that Terraform is here to reflect what's the current state. So when I specify `1.25.6` in Terraform, I expect AKS to be on `1.25.6` and not conflict with each-other, unless you explicitly told it to upgrade itself in the background, but then you are responsible for updating your HCL code as well (this module sort of helps you there as it has checks that prevent you from configuring automatic patch upgrades if k8s patch versions are specified).

The problem for us (and probably other companies) is that we have a pinned version of the Terraform code for most of our clusters which get's only updated during regular releases. So we can't just change the code whenever AKS updated to a new patch version. That's why we only specify `1.25` in our code. So then this is a legitimate problem that Terraform shouldn't be confused when the AKS API returns `1.25.6` instead of `1.25` as it's outside of what Terraform should manage for us.

Does this explain some of our background?

Of course it's up to discussion who should handle that drift and as already discussed in this PR the ideal situation would be if the AKS API doesn't return a patch version (or in a separate field) when we haven't specified one in Terraform. 

Thus I think this PR is a good workaround but shouldn't be considered a final implementation right? The final implementation would be done in the AzureRM provider as soon as the AKS API changes it's spec?

But for the meantime I could live with this workaround and as @viters already said it's better to have one rather than being blocked or eventually letting Terraform downgrade/recreate your cluster out of a sudden.

Oh and one last side-node: If you are still on `1.24` as we are, this problem doesn't affect you that much as there are not that frequent (if any) new patches for `1.24` ;)

the-technat:
@zioproto what sort of feedback should we add to the related AKS API issue?

zioproto:
> @zioproto what sort of feedback should we add to the related AKS API issue?

+1 the issue https://github.com/Azure/AKS/issues/3573 with a 👍 

I see that you wrote on this thread:

>  this is a legitimate problem that Terraform shouldn't be confused when the AKS API returns 1.25.6 instead of 1.25

This is a good feedback to report on that issue.

cveld:
Not sure why it happens, but we are on 1.24 and even then after a while aks has changed the property to 1.24.9 and the azurerm provider detects a change. The azurerm provider does not recreate the cluster, but it does take its time to propagate the change from 1.24.9 to 1.24. This change detection is undesired.

the-technat:
> Not sure why it happens, but we are on 1.24 and even then after a while aks has changed the property to 1.24.9 and the azurerm provider detects a change. The azurerm provider does not recreate the cluster, but it does take its time to propagate the change from 1.24.9 to 1.24. This change detection is undesired.

Exact same behavior here, we're on 1.24 too and nothing happened for weeks until suddenly this morning the change happened (e.g the clusters are updating and reporting the wrong version).

lonegunmanb:
I'd like to merge this pr as a mitigation method, we have about one month to make the final decision, any objection?

the-technat (approved):

jiaweitao001 (approved):
LGTM.

main.tf
swapnil-gohil: @the-technat So we're currently upgrading our AKS module from 6.2.0 --> 7.2.0. We're getting a module.aks.azapi_update_resource.aks_cluster_post_create to be added. Can you please make me understand what is the purpose the resource created and is it safe to apply? Will the future updates of AKS take place through the created resource? Also we don't have azapi provider in our code is it necessary to add? Thanks.

PR #357, opened 2023-04-25
feat!: add create_before_destroy=true to node pools

### LINKED ISSUE #358: Support for create_before_destroy on node pools

### Description

If Terraform ever has to recreated a node pool, I want this to happen without any down-time for applications. In order to prevent nodes from being recreated, Terraform supports a [lifecycle argument](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#create_before_destroy) `create_before_destroy`. With this, a new node pools with new nodes is added first, before the old nodes are removed ensuring service continuity.

I'd like the AKS module to support this lifecycle argument on node pools.

azurerm_kubernetes_cluster_node_pool

```hcl
resource "azurerm_kubernetes_cluster_node_pool" "node_pool" {
  [...]

  lifecycle {
    create_before_destroy = true
  }
}
```

Please note that the lifecycle block can't be filled with variables. There's no way to make this setting configurable (I tried with a dynamic block and with variables both in and outside of the node_pool var).

In addition, node pool names have to be unique, so a random suffix or so would be required too, to avoid name clashes. And then that would generate a breaking change.

### References

PR: https://github.com/Azure/terraform-azurerm-aks/pull/357

the-technat:
## Describe your changes

While migrating our node pools into the module, we noticed that the lifecycle attribute `create_before_destroy` is not set.

This is an attribute that should be set to true by default to allow for rolling-updates of the node pools whenever there are changes to them.

Please note that Terraform forbids adding a variable in the lifecycle block or even setting the lifecycle block to dynamic, that's why this value is **hard coded**, but it shouldn't harm anyone and [terraform-aws-eks also has this hardcoded](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/modules/eks-managed-node-group/main.tf#L381).

## Issue number
Closes #358 

Thanks for your cooperation!


main.tf
```diff
@@ -592,6 +592,8 @@ resource "azurerm_kubernetes_cluster_node_pool" "node_pool" {
   depends_on = [azapi_update_resource.aks_cluster_post_create]
 
   lifecycle {
+    create_before_destroy = true
```
mkilchhofer, line 595: At least README notes are required if we "hardcode" this. Nodepool names need to be unique and `create_before_destroy = true` will create another nodepool with the same name by default.

the-technat: Jep definitely. Since the change is hard coded (and can only be hard coded), I think we should also hard coded a random suffix to the node pool?

lonegunmanb: Yeah I like this random suffix idea.

Btw, it's possible to make this `create_before_destroy = true` dynamically, we can declare two similar resources with a toggle switcher, one contains `create_before_destroy = true` and another doesn't. Like:

```hcl
resource "azurerm_kubernetes_cluster_node_pool" node_pool {
  count = var.create_before_destroy ? 0 : 1

 ...
}

resource "azurerm_kubernetes_cluster_node_pool" node_pool_create_before_destroy {
  count = var.create_before_destroy ? 1 : 0

 ...
  create_before_destroy = true
}
```

But in this case, we won't need this trick, I agree with you, the node pool should contain `create_before_destroy`.

I have another question @the-technat, even with `create_before_destroy = true` how can we achieve seamless node pool updates? The old pool would be destroyed immediately once the new pool is ready, would it cause downtime?

Another option is we maintain the node pool in green-blue pattern, the upgrade could create a blue pool, then drain the green pool, then destroy the green pool. I would consider it as plan B if re-creation of a node pool with `create_before_destroy` still cause downtime.

lonegunmanb: I'm thinking the current implementation of using `random_string`'s result won't be refreshed on the node pool's recreation.

the-technat: You're right! And since the random_string can't depend on the node pool for lifecycle triggers we need to solve this otherwise.

Changed it to how we handle this currently on our managed node pools, if the configuration changes in any way, the name will change since the hash isn´t the same anymore. I'm just thinking if all parameters on the node pool should trigger it or if there are certain fields that don't need a replacement.

the-technat: Hm but we don't only need a new string when the name changes, but evertime a setting is changed and the nodepool is replaced.

lonegunmanb: > Hm but we don't only need a new string when the name changes, but evertime a setting is changed and the nodepool is replaced.

Since the suffix is calculated from `substr(md5(jsonencode(each.value)` so every time the pool is recreated, a new name would be generated, because the recreation must be triggered by some attribute's change.

the-technat: Okay but that's more or less what is currently proposed in the PR. What I still don't get is why you need the null-resource to trigger a replacement based on the name? The name is within `each.value` and would change the hash too, right?

zioproto: > Regarding your other question, that totally depends on how the AzureRM API implements deletion of the old node pool. I assume it starts with cordening all the nodes and then draining them, which results in workload beeing moved to the new nodes. That should give you a seamless update. And in order to ensure total 100% availability of the apps, they have to use [PDBs](https://kubernetes.io/docs/tasks/run-application/configure-pdb/) (which they should use anyway), to prevent K8s from draining nodes too fast (draining will respect PDBs and wait until the pods are started on the new nodes).

That is correct. When the API call `az aks nodepool delete` is used, the nodes are cordoned and drained, and when the workloads are gone the VMs are finally deleted. I actually could not find a confirmation in the docs, so I tested this behaviour to make sure it is correct.

the-technat: Got it, but why did we added name to `ignore_changes`? Wouldn't without that the `null_resource` be obsolete?

lonegunmanb: > Got it, but why did we added name to `ignore_changes`? Wouldn't without that the `null_resource` be obsolete?

`name                          = "${each.value.name}-${substr(md5(jsonencode(each.value)), 0, 4)}"`

Any change in `each.value` would cause a new name suffix, and `name` is a `ForceNew` attribute so any change would cause a recreation on node pool, so we have to put `name` in `ignore_changes`. A `null_resource` would guarantee that once we do change the name, the pool would be recreated.

the-technat: And what if I change the instance type?

the-technat: Ah... now finally I get it. Jep would make sense to do some tests with your proposed code.

lonegunmanb: Yeah, we do need sufficient tests for this Terraform circus show 😃

lonegunmanb: Hi @the-technat, I've done some tests on my side and the results are good, I've tried updating name, force new attributes and update-in-place attributes, and all scenarios work as we expected. I would encourage you to do the same tests on your side so we can on the same page.

the-technat: @lonegunmanb I did some tests with your suggestion and it worked fine. Now I also see why my apporach wouldn't work (it recreates the nodepool for every non-breaking attribute too...).

I updated the PR with your approach and fixed one test (now nodepool names can only be up to 8 chars since the module itself adds 4). Still not very happy with the `null_resource`, but I guess that's our only option. Maybe it's worth creating a feature request for the AKS API to implement a new field `use_name_prefix=true` on the nodepool that will automatically use your given name as prefix and inject a random suffix if the node pool get's recreated.

lonegunmanb: @the-technat Terraform 1.4 has added a new built-in resource [`terraform_data`](https://www.infoq.com/news/2023/03/terraform-1-4-null-resource/) to replace `null_resource`, so I'm thinking of whether should we use this new resource or not. Using this new resource could set us free from `null` provider, but it also require the Terraform Core's version must be `>= 1.4`.

This `null_resource` + `md5` + `replace_triggered_by` pattern can be used in multiple scenarios, and I think pulumi has faced the very same problem so they've introduced [auto-naming](https://www.pulumi.com/docs/intro/concepts/resources/names/#autonaming).

the-technat: Interesting, `terraform_data` definitely looks like a better approach.

Because this is a breaking change anyway I guess we could do this. Otherwise I'd be afraid of bumping the Terraform version to `>= 1.4` as this has affect on all parent / sibling-modules.

lonegunmanb (approved):
Thanks @the-technat, LGTM! I'm merging this pr now, as for the `terraform_data` proposal, we can open a separate pr for it.

Issue #364, opened 2023-05-12
The given "for_each" argument value is unsuitable: "for_each" sets must not contain null values when using create_role_assignment_network_contributor


avekrivoy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Greenfield/Brownfield provisioning

greenfield

### Terraform Version

1.4.6

### Module Version

6.8.0

### AzureRM Provider Version

3.56.0

### Affected Resource(s)/Data Source(s)

azurerm_role_assignment, azurerm_kubernetes_cluster_node_pool

### Terraform Configuration Files

```hcl
locals {
  node_pools = {
    monitoring = {
      name            = "monitoring"
      node_count      = 1
      vm_size         = "standard_d2ads_v5"
      max_pods        = 250
      node_labels = {
        "workload_type" = "monitoring"
      }
    }
  }
}

module "aks-cluster" {
  source                                     = "Azure/aks/azurerm"
  version                                    = "6.8.0"
  resource_group_name                        = azurerm_resource_group.rg.name
  location                                   = azurerm_resource_group.rg.location
  kubernetes_version                         = var.kubernetes_version
  orchestrator_version                       = var.kubernetes_version
  private_cluster_enabled                    = var.private_cluster_enabled
  prefix                                     = local.aks_prefix

  agents_pool_name                           = "system"
  agents_size                                = var.system_nodepool_agents_size
  agents_count                               = var.system_nodepool_agents_count
  agents_max_pods                            = var.system_nodepool_max_pods
  agents_availability_zones                  = var.system_nodepool_agents_availability_zones
  agents_labels                              = var.system_nodepool_agents_labels
  only_critical_addons_enabled               = var.only_critical_addons_enabled
  role_based_access_control_enabled          = var.role_based_access_control_enabled
  rbac_aad                                   = var.rbac_aad
  create_role_assignment_network_contributor = var.create_role_assignment_network_contributor

  network_plugin                             = var.network_plugin
  vnet_subnet_id                             = local.aks_vnet_subnet_id
  net_profile_service_cidr                   = var.net_profile_service_cidr
  net_profile_dns_service_ip                 = var.net_profile_dns_service_ip
  attached_acr_id_map                        = local.attached_acr_id_map
  maintenance_window                         = var.maintenance_window
  public_ssh_key                             = var.public_ssh_key
  # Addons
  key_vault_secrets_provider_enabled         = var.key_vault_secrets_provider_enabled
[...]

PR #374, opened 2023-05-23
Filter null value out from `local.subnet_ids`

lonegunmanb:
## Describe your changes

As #364 described, sometimes `local.subnet_ids` would contain `null` value, and that could cause error as the set is used as `for_each` iterator. This pr filtered `null` value out of the set to solve the issue.

## Issue number

#364 

Thanks for your cooperation!


jiaweitao001 (approved):
🚢

PR #395, opened 2023-06-16
Fix incorrect e2e test code so it could pass on our local machine

lonegunmanb:
## Describe your changes

The e2e test `TestExamples_differentLocationForLogAnalyticsSolution` would failed when we run e2e test on our local machine because of the incorrect initialization. This pr fixed this incorrect e2e test.

## Issue number

#000

Thanks for your cooperation!

### DIFF
--- a/test/e2e/terraform_aks_test.go
+++ b/test/e2e/terraform_aks_test.go
@@ -38,7 +38,7 @@ func assertOutputNotEmpty(t *testing.T, output test_helper.TerraformOutput, name
 }
 
 func TestExamplesWithoutMonitor(t *testing.T) {
-	var vars map[string]interface{}
+	vars := make(map[string]interface{}, 0)
 	managedIdentityId := os.Getenv("MSI_ID")
 	if managedIdentityId != "" {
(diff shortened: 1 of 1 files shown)


zioproto:
LGTM, I tested and everything works as expected:
```
PASS
ok      github.com/Azure/terraform-azurerm-aks/e2e      2541.921s
```

zioproto:
The E2E test in GitHub failed because of https://www.ipify.org/ returning Bad Gateway

```
│     │ while calling jsondecode(str)
│     │ data.curl.public_ip[0].response is "Bad Gateway"
```

lonegunmanb:
> The E2E test in GitHub failed because of https://www.ipify.org/ returning Bad Gateway
> 
> ```
> │     │ while calling jsondecode(str)
> │     │ data.curl.public_ip[0].response is "Bad Gateway"
> ```

I'm thinking of composing a new small module that returns your current public ip, and in this new module we should retrieve public ip from multiple APIs in case one is not available temporarily.

jiaweitao001 (approved):
🚢

Issue #411, opened 2023-07-17
Log analytics workspaces was not found


ghost:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Greenfield/Brownfield provisioning

greenfield

### Terraform Version

1.4.6

### Module Version

7.2.0

### AzureRM Provider Version

3.62.1

### Affected Resource(s)/Data Source(s)

azurerm_log_analytics_workspace

### Terraform Configuration Files

```hcl
terraform {
  required_providers {
    azurerm = {
      source  = "hashicorp/azurerm"
      version = ">=3.0.0"
    }
  }
}

variable "subscription_id_1" {}
variable "subscription_id_2" {}

provider "azurerm" {
  features {}

  subscription_id = var.subscription_id_1
}

provider "azurerm" {
  alias = "log"

  features {}

  subscription_id = var.subscription_id_2
}

resource "azurerm_resource_group" "log" {
  name     = "log-347347347"
  location = "West Europe"

  provider = azurerm.log
}

resource "azurerm_log_analytics_workspace" "log" {
  name                = "log-856785685468"
  location            = azurerm_resource_group.log.location
  resource_group_name = azurerm_resource_group.log.name

  provider = azurerm.log
}

resource "azurerm_resource_group" "aks" {
  name     = "aks-34574357457"
  location = "West Europe"
}

module "aks" {
  source                          = "Azure/aks/azurerm"
  version                         = "7.2.0"
  resource_group_name             = azurerm_resource_group.aks.name
  prefix                          = "425867245986"
  rbac_aad                        = false
  log_analytics_workspace_enabled = true
  log_analytics_workspace = {
    id   = azurerm_log_analytics_workspace.log.id
    name = azurerm_log_analytics_workspace.log.name
  }

  depends_on = [azurerm_resource_group.aks]
}
```

### tfvars variables values

```hcl
subscription_id_1 = ""
subscription_id_2 = ""
```

### Debug Output/Panic Output

```shell
╷
│ Error: log analytics workspaces "log-856785685468" (Resource Group "log-347347347") was not found
│ 
│   with module.aks.data.azurerm_log_analytics_workspace.main[0],
│   on .terraform/modules/aks/main.tf line 725, in data "azurerm_log_analytics_workspace" "main":
│  725: data "azurerm_log_analytics_workspace" "main" {
│ 
╵
```

### Expected Behaviour

[...]

PR #412, opened 2023-07-20
Add `location` and `resource_group_name` for `var.log_analytics_workspace`

lonegunmanb:
Add `location` and `resource_group_name` for `var.log_analytics_workspace` so users can set these two fields and avoid
data.azurerm_log_analytics_workspace in case they'd like to use resource in another subscription.

## Issue number

#411

Thanks for your cooperation!


jiaweitao001 (approved):
🚢

Issue #421, opened 2023-08-09
Support for confidental computin


xWhiteTonicx:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Confidental compution support for use Intel SGX. It had been already addd to terraform provider. 

### New or Affected Resource(s)/Data Source(s)

azurerm_kubernetes_cluster

### Potential Terraform Configuration

```hcl
variable "enable_confidential_computing" {
  type        = bool
  default     = false
  description = "(Optional) Should confidential compution be enabled?"
}

resource "azurerm_kubernetes_cluster" "main" {
   [..CUT..]
  dynamic "confidential_computing" {
    for_each = var.enable_confidential_computing ? ["confidential_computing"] : []
    content {
      sgx_quote_helper_enabled = true
    }
  }
   [..CUT..]
}
```

### References

https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster#confidential_computing
https://learn.microsoft.com/en-us/azure/confidential-computing/confidential-nodes-aks-overview

lonegunmanb:
I don't think this feature is a preview feature, so bring it on!

PR #423, opened 2023-08-10
Add confidential computing in aks module

jiaweitao001:
## Issue number

#[421](https://github.com/Azure/terraform-azurerm-aks/issues/421)

Thanks for your cooperation!


variables.tf
```diff
@@ -388,6 +388,14 @@ variable "cluster_name" {
   description = "(Optional) The name for the AKS resources created in the specified Azure Resource Group. This variable overwrites the 'prefix' var (The 'prefix' var will still be applied to the dns_prefix if it is set)"
 }
 
+variable "confidential_computing" {
+  type = list(object({
```
lonegunmanb, line 392: According to the [schema](https://github.com/hashicorp/terraform-provider-azurerm/blob/main/internal/services/containers/kubernetes_addons.go#L89-L101), could we set type to `boolean` for this variable?

lonegunmanb (changes requested):
Thanks @jiaweitao001, one comment.

variables.tf
jiaweitao001: Yes, we can simplify this type to a `boolean`, I was wondering there might be a possibility that more attributes will be added to this `confidential_computing` in the future, use a `list` would be more scalable?

lonegunmanb: Since the `MaxItems` of this list is `1` I think in that case, an `object` would be enough for us. WDYT?

jiaweitao001: Make sense. We can just use `object` instead of `list`.

lonegunmanb (approved):
Thanks @jiaweitao001, LGTM!

Issue #438, opened 2023-09-07
Support for custom_ca_trust_enabled in the node_pools


isantospardo:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

We would like to be able to enable custom_ca_trust_enabled within this module described in the upstream Terraform Kubernetes Cluster [resource](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster).

This feature is currently not implemented yet and it would let us configure the custom_ca, it is known it is in a [preview state](https://learn.microsoft.com/en-us/azure/aks/custom-certificate-authority)  but it is already integrated into the latest available aks cluster provider

### New or Affected Resource(s)/Data Source(s)

azurerm_kubernetes_cluster

### References

https://learn.microsoft.com/en-us/azure/aks/custom-certificate-authority
https://github.com/claranet/terraform-azurerm-aks
https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster

zioproto:
Feature still in preview:
https://github.com/Azure/AKS/issues/2259

Probably this issue is a duplicated of #379

lonegunmanb:
I'm closing this issue since the feature has been deprecated.

PR #439, opened 2023-09-07
feat(custom-ca-trust): Enable custom_ca_trust_enabled in node_pools

isantospardo:
## Describe your changes

We would like to be able to enable custom_ca_trust_enabled within this module described in the upstream Terraform Kubernetes Cluster [resource](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster).

This feature is currently not implemented yet and it would let us configure the custom_ca, it is known it is in a [preview state](https://learn.microsoft.com/en-us/azure/aks/custom-certificate-authority) but it is already integrated into the latest available aks cluster provider

## Issue number
 
#438 

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -52,6 +52,7 @@ resource "azurerm_kubernetes_cluster" "main" {
     content {
       name                         = var.agents_pool_name
       vm_size                      = var.agents_size
+      custom_ca_trust_enabled      = var.custom_ca_trust_enabled
       enable_auto_scaling          = var.enable_auto_scaling
       enable_host_encryption       = var.enable_host_encryption
       enable_node_public_ip        = var.enable_node_public_ip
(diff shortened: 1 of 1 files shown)


lonegunmanb:
Thanks @isantospardo for opening this pr. For now, we won't merge the preview features since the service team might revoke them at any time, so I'd like to wait until this feature is GA.

sophher:
Looks like this feature is finally in GA
https://github.com/Azure/AKS/issues/2259

lonegunmanb:
@sophher but `custom_ca_trust_enabled` has been removed in `v4`: https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/guides/4.0-upgrade-guide#custom-ca-trust

sophher:
@lonegunmanb as far as I can see, it was removed in v4, because it was only in preview back then and with v4 terraform removed all preview features.

> In order to provide a more reliable experience when using the Azure Provider, Microsoft has requested that the AKS resource only use a stable API version. Starting in version 4.0 of the AzureRM provider, Microsoft and HashiCorp have agreed to switch to the AKS stable APIs. This means that some preview features that are exposed in 3.x will no longer be available in 4.x. The following properties will no longer be available in 4.x:

Now it is in GA and it should be possible to use it again via Terraform.

PR #459, opened 2023-10-26
Include the `azurerm_kubernetes_cluster` identity block in the module's output

mr-onion-2:
Added the `azurerm_kubernetes_cluster`'s identity block to the module's output

Thanks for your cooperation!


zioproto:
@mr-onion-2 there is already an output `cluster_identity`. Can you explain why we need this additional output ? thanks

mr-onion-2:
Oh so there is! 
I did see that but I assumed it was some other identity - I should have noticed it is indeed the same identity block

Thanks, I'll close

Issue #429, opened 2023-08-24
network_plugin_mode expects "Overlay" but azurerm provider expect "overlay" resulting in constant cluster destroy


daximillian:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Greenfield/Brownfield provisioning

greenfield

### Terraform Version

1.5.4

### Module Version

7.2.0

### AzureRM Provider Version

3.64.0

### Affected Resource(s)/Data Source(s)

azurerm_kubernetes_cluster

### Terraform Configuration Files

```hcl
network_plugin                    = "azure"
  network_plugin_mode               = "Overlay"    
  ebpf_data_plane                   = "cilium"
```

### tfvars variables values

```hcl
N/A
```

### Debug Output/Panic Output

```shell
network_profile {
          + network_mode        = (known after apply)
          ~ network_plugin_mode = "overlay" -> "Overlay" # forces replacement
          ~ network_policy      = "cilium" -> (known after apply)
          ~ pod_cidr            = "10.244.0.0/16" -> (known after apply)
          ~ pod_cidrs           = [
              - "10.244.0.0/16",
            ] -> (known after apply)
```

### Expected Behaviour

Not have the capitalization of the network_plugin_mode force a replacement.

### Actual Behaviour

Setting the network_plugin_mode to "Overlay" forces a replacement, and setting it to "overlay" results in an error.

zioproto:
hello @daximillian ,
this is a provider issue, and should be opened on the repo https://github.com/hashicorp/terraform-provider-azurerm

I checked, and the issue has already been fixed with provider version **v3.66.0**
https://github.com/hashicorp/terraform-provider-azurerm/issues/22621

I see you are using provider version 3.64.0, please upgrade

zioproto:
Should we upgrade the requirement to v3.66.0 if using Overlay ?

https://github.com/Azure/terraform-azurerm-aks/blob/0833586b9a22b683fcd1a774372956ae8f9da5ba/versions.tf#L9-L12

lonegunmanb:
> Should we upgrade the requirement to v3.66.0 if using Overlay ?
> 
> https://github.com/Azure/terraform-azurerm-aks/blob/0833586b9a22b683fcd1a774372956ae8f9da5ba/versions.tf#L9-L12

Good idea!

lonegunmanb:
Hi

> Should we upgrade the requirement to v3.66.0 if using Overlay ?
> 
[...]

PR #472, opened 2023-11-10
use lowercase everywhere for network plugin mode overlay

zioproto:
## Describe your changes

Using `Overlay` instead of `overlay` triggers unnecessary cluster updates.
Fixing to user lowercase only everywhere.

## Issue number
https://github.com/Azure/terraform-azurerm-aks/issues/429 
https://github.com/hashicorp/terraform-provider-azurerm/issues/22621


lonegunmanb (approved):
Thanks @zioproto, LGTM!

PR #481, opened 2023-11-28
Add web_app_routing_identity block to outputs

bonddim:
## Describe your changes
This add `web_app_routing_identity` to outputs, which is needed for role assignments when DNS zone and AKS Cluster are in different resource groups.

### DIFF
--- a/outputs.tf
+++ b/outputs.tf
@@ -223,3 +223,8 @@ output "username" {
   sensitive   = true
   value       = azurerm_kubernetes_cluster.main.kube_config[0].username
 }
+
+output "web_app_routing_identity" {
+  description = "The `azurerm_kubernetes_cluster`'s `web_app_routing_identity` block."
+  value       = try(azurerm_kubernetes_cluster.main.web_app_routing[0].web_app_routing_identity, null)
+}
--- a/outputs.tf
+++ b/outputs.tf
@@ -225,6 +225,6 @@ output "username" {
 }
 
 output "web_app_routing_identity" {
-  description = "The `azurerm_kubernetes_cluster`'s `web_app_routing_identity` block."
-  value       = try(azurerm_kubernetes_cluster.main.web_app_routing[0].web_app_routing_identity, null)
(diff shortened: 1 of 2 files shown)


zioproto (approved):

outputs.tf
```diff
@@ -223,3 +223,8 @@ output "username" {
   sensitive   = true
   value       = azurerm_kubernetes_cluster.main.kube_config[0].username
 }
+
+output "web_app_routing_identity" {
+  description = "The `azurerm_kubernetes_cluster`'s `web_app_routing_identity` block."
+  value       = try(azurerm_kubernetes_cluster.main.web_app_routing[0].web_app_routing_identity, null)
```
lonegunmanb, line 229: According to the [schema](https://github.com/hashicorp/terraform-provider-azurerm/blob/main/internal/services/containers/kubernetes_cluster_resource.go#L495-L514), `web_app_routing_identity` is a list of object, so I would recommend `[]` as default value instead of `null`, so the consumer can use `for` expression to iterate it:

```hcl
value       = try(azurerm_kubernetes_cluster.main.web_app_routing[0].web_app_routing_identity, [])
```

Otherwise, we'll receive an error when we're using `for` expression and this output is `null`:

```text
Error: Iteration over null value
```

outputs.tf
```diff
@@ -223,3 +223,8 @@ output "username" {
   sensitive   = true
   value       = azurerm_kubernetes_cluster.main.kube_config[0].username
 }
+
+output "web_app_routing_identity" {
+  description = "The `azurerm_kubernetes_cluster`'s `web_app_routing_identity` block."
```
lonegunmanb, line 228: I'd like to address that this block is a list of object:

description = "The `azurerm_kubernetes_cluster`'s `web_app_routing_identity` block, it's type is a list of object."

lonegunmanb (changes requested):
Thanks @bonddim for opening this pr! Almost LGTM but two review comments.

outputs.tf
bonddim: Done

outputs.tf
bonddim: Done

PR #435, opened 2023-08-29
feat(http_proxy_config): Add http_proxy_config

isantospardo:
## Describe your changes

Adds a HTTP(S) proxy configuration within this module like described in the upstream Terraform Kubernetes Cluster [resource](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster).

This feature is currently not implemented yet and it would let us configure a HTTP(s) proxy for our AKS nodes with a customer CA as well.

You can have a look on a fork which enabled it, please see [here](https://github.com/claranet/terraform-azurerm-aks/blob/master/README.md?plain=1#L299).

We are working on a PR already and it will be linked to this issue asap.

Thanks for your understanding and time.
## Issue number

#434 

## Checklist before requesting a review
- [x] The pr title can be used to describe what this pr did in `CHANGELOG.md` file
- [x] I have executed pre-commit on my machine
- [x] I have passed pr-check on my machine

Fixes: https://github.com/Azure/terraform-azurerm-aks/issues/434

Issue #434, opened 2023-08-29
Support for http proxy settings


vmaillot:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

We would like to be able to setup a HTTP(S) proxy configuration within this module like described in the upstream Terraform Kubernetes Cluster [resource](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster).

This feature is currently not implemented yet and it would let us configure a HTTP(s) proxy for our AKS nodes with a customer CA as well. 

You can have a look on a fork which enabled it, please see [here](https://github.com/claranet/terraform-azurerm-aks/blob/master/README.md?plain=1#L299).

We already opened a [PR](https://github.com/Azure/terraform-azurerm-aks/pull/435).

Thanks for your understanding and time.
 

### New or Affected Resource(s)/Data Source(s)

azurerm_kubernetes_cluster

### References

https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster
https://github.com/claranet/terraform-azurerm-aks

zioproto:
Blocked by https://github.com/hashicorp/terraform-provider-azurerm/issues/24188

JiteshShivnani:
How can I setup the proxy in terraform for azure? The following is not working. It says http_proxy_config in main file is not allowed.

In my **variables.tf**

```
variable "http_proxy_config" {
	type = object({
		http_proxy = optional(string)
		https_proxy = optional(string)
		no_proxy = optional(list(string))
		trusted_ca = optional(string)
		})
	default = {
		http_proxy = "http://*******.com:8080"
		https_proxy = "http://****.com:8080"
#		no_proxy = 
#		trusted_ca =
 }
 }
```

In my **main.tf**

```
module "aks" {
    source = "../../../modules/aks"
    cluster_name = var.cluster_name
    dns_prefix = var.dns_prefix
    private_cluster_enabled = var.private_cluster_enabled
    private_dns_zone_id = var.private_dns_zone_id
    default_node_pool_name = var.default_node_pool_name
    node_pool_count = var.node_pool_count
    node_pool_vm_size = var.node_pool_vm_size
    vnet_subnet_id = data.azurerm_subnet.readysn.id
    node_pool_enable_auto_scaling = var.node_pool_enable_auto_scaling
    node_pool_autoscaling_max_count = var.node_pool_autoscaling_max_count
[...]

PR #492, opened 2023-12-20
feat(`http_proxy_config`): Add `http_proxy_config`

lonegunmanb:
## Describe your changes

It's a forked pr based on #435, @isantospardo should take the credit. It has been a long time since #435 was opened and a lot of code have been changed, so I just fork #435's branch and rebase it with the latest `main` branch.

Adds a HTTP(S) proxy configuration within this module like described in the upstream Terraform Kubernetes Cluster [resource](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster).

This feature is currently not implemented yet and it would let us configure a HTTP(s) proxy for our AKS nodes with a customer CA as well.

You can have a look on a fork which enabled it, please see [here](https://github.com/claranet/terraform-azurerm-aks/blob/master/README.md?plain=1#L299).

## Issue number

#434 

Thanks for your cooperation!


main.tf
```diff
@@ -584,6 +595,12 @@ resource "azurerm_kubernetes_cluster" "main" {
   }
 }
 
+resource "null_resource" "aks_cluster_recreate" {
```
zioproto, line 598: I don't understand this name ? Is it attempting to recreate the all cluster if the noproxy value changes ?

I dont see the corresponding `azapi_update_resource` with the `lifecycle.replace_triggered_by` section. Is it missing ?

zioproto (changes requested):

main.tf
lonegunmanb: You're correct, I found that we can update this property instead of create the whole cluster. I'll update this pr.

variables.tf
```diff
@@ -57,6 +57,24 @@ variable "agents_min_count" {
...
+    http_proxy  = optional(string)
+    https_proxy = optional(string)
+    no_proxy    = optional(list(string))
+    trusted_ca  = optional(string)
+  })
+  default     = null
+  description = <<-EOT
+    optional(object({
+      http_proxy  = (Optional) The proxy address to be used when communicating over HTTP. Changing this forces a new resource to be created.
+      https_proxy = (Optional) The proxy address to be used when communicating over HTTPS. Changing this forces a new resource to be created.
```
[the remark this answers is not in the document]
lonegunmanb: Thanks for sharing! In this case, I'd like to follow the same behavior, once you've set one of them but not both, I'll set them both for the user to avoid the configuration drift.

zioproto (approved):

Issue #491, opened 2023-12-19
Concern about random suffix on nodepools in version 7


tdevopsottawa:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

I'm planning an upgrade to version 7 and from this:

```
Now azurerm_kubernetes_cluster_node_pool.node_pool resource has create_before_destroy=true to avoid downtime when upgrading node pools. Users must be aware that there would be a "random" suffix added into pool's name, this suffix's length is 4, so your previous node pool's name nodepool1 would be nodepool1xxxx. This suffix is calculated from node pool's config, the same configuration would lead to the same suffix. You might need to shorten your node pool's name because of this new added suffix.

To enable this feature, we've also added new null_resource.pool_name_keeper to track node pool's name in case you've changed the name.
```

It sounds like the new version of the module adds a random suffix to the end of every node pool name. This is a very undesirable thing for my environment as we've got a huge CI stack set up where kubernetes manifests across lots of clusters are being set to deploy to specific node pool names with the node pool names hardcoded in the manifests.

If my understanding is correct is there some way to turn off this feature and just keep the simple node pool names? If there isn't this basically blocks me from upgrading. 

### New or Affected Resource(s)/Data Source(s)

azurerm_kubernetes_cluster

zioproto:
Related to:
https://github.com/Azure/terraform-azurerm-aks/issues/476

zioproto:
Hello @tdevopsottawa 

I understand you are using the default node label `kubernetes.azure.com/agentpool` to deploy Pods to specific node pool names using `nodeSelector`.

my suggestion is to use your specific node labels for the Kubernetes scheduler:
https://github.com/Azure/terraform-azurerm-aks/blob/6e2f254dca9a23e2b143426b1e639af245304789/variables.tf#L889

@mkilchhofer @the-technat you folks originally contributed to https://github.com/Azure/terraform-azurerm-aks/pull/357
Do you have any suggestion here ?

thanks

the-technat:
@tdevopsottawa I assume you are using multiple node pools with different node sizes / configurations that you want to schedule on?

To give some background: we initially wanted to implement this behavior in a non-breaking way, but Terraform [doesn't allow you to specify lifecycle arguments dynamically](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#literal-values-only).

[...]

Issue #389, opened 2023-06-08
terraform should destroy and create resources if there is change in settings of aks cluster 


nayaksuraj:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Greenfield/Brownfield provisioning

brownfield

### Terraform Version

1.4.6

### Module Version

7.1.0

### AzureRM Provider Version

3.58.0

### Affected Resource(s)/Data Source(s)

azurerm_kubernetes_cluster

### Terraform Configuration Files

```hcl
resource "random_id" "prefix" {
  byte_length = 8
}

resource "azurerm_resource_group" "main" {
  count = var.create_resource_group ? 1 : 0

  location = var.location
  name     = coalesce(var.resource_group_name, "${random_id.prefix.hex}-rg")
}

locals {
  resource_group = {
    name     = var.create_resource_group ? azurerm_resource_group.main[0].name : var.resource_group_name
    location = var.location
  }
}

resource "azurerm_virtual_network" "test" {
  address_space       = ["10.52.0.0/16"]
  location            = local.resource_group.location
  name                = "${random_id.prefix.hex}-vn"
  resource_group_name = local.resource_group.name
}

resource "azurerm_subnet" "test" {
  address_prefixes                               = ["10.52.0.0/24"]
  name                                           = "${random_id.prefix.hex}-sn"
  resource_group_name                            = local.resource_group.name
  virtual_network_name                           = azurerm_virtual_network.test.name
  enforce_private_link_endpoint_network_policies = true
}

locals {
  nodes = {
    for i in range(1) : "w${i}" => {
      name           = substr("worker${i}${random_id.prefix.hex}", 0, 8)
      vm_size        = "Standard_D2s_v3"
      node_count     = 1
      vnet_subnet_id = azurerm_subnet.test.id
    }
  }
}

module "aks" {
  source = "../.."

  prefix                        = "prefix-${random_id.prefix.hex}"
  resource_group_name           = local.resource_group.name
  os_disk_size_gb               = 50
  public_network_access_enabled = true
  sku_tier                      = "Standard"
  rbac_aad                      = false
  vnet_subnet_id                = azurerm_subnet.test.id
  node_pools                    = {}
  agents_pool_name = "np"
}
```

### tfvars variables values

```hcl
variable "create_resource_group" {
  type     = bool
  default  = false
  nullable = false
}

variable "location" {
  default = "westeurope"
}

variable "resource_group_name" {
  type    = string
[...]

PR #496, opened 2024-01-11
Let the users decide whether adding a random suffix in cluster and pool's name or not.

lonegunmanb:
As [#389](https://github.com/Azure/terraform-azurerm-aks/issues/389#issuecomment-1670748181) described, now `azurerm_kubernetes_cluster.main` has been set to `create_before_destroy = true` implicity because an `azurerm_kubernetes_cluster_node_pool` with `create_before_destroy = true` referred it, this meta argument config has been propagated to the aks resource.

We also have #491 complaining that they don't want a random suffix in node pool's name. We added `create_before_destroy = true` to the node pool resource in #357 since we'd like to create a new replacement pool before we destroy the old one so all running pods on the old pool could be evicted to the new one before the worker node have been destroyed, so the cluster's operator could avoid downtime when they have to recreate the node pool.

We respect that the users should have the right to decide the style they'd adapted already, so this pr replicated a new `azurerm_kubernetes_cluster_node_pool`, and added `create_before_destroy` in `var.node_pools` object definition.

Setting this `create_before_destroy` to `true`, nothing would happen, your node pool would still be created and managed in the old way, with a random name suffix.

Setting this `create_before_destroy` to `false`, the node pool in brown field would be destroyed and recreated in new address, without random name suffix, and when Terraform have to recreate this node pool, you must delete the old pool and create the new replacement pool with a different name. To avoid downtime we encourage the operators to create the replacement pool first, then destroy the old one.

We also added a new variable `cluster_name_random_suffix`, setting this variable to `true` would add a random suffix to cluster's name when you're creating an Aks instance in greenfield, but will do nothing to the instances in brownfield. With this random suffix now Terraform could recreate the Aks instance directly without any naming conflict.

@zioproto @TechieNat @tdevopsottawa WDYT?

## Issue number

#000

Thanks for your cooperation!


zioproto:
I believe you meant to tag @the-technat ☝️

the-technat (approved):
It's a prone to errors doing it that way, but that should certainly give the user the flexibility to do it the way he wants.

Thanks @lonegunmanb for coding this!

zioproto (commented):
It works. I tested like this:

```
locals {
  node_pools = {
    user = {
      name                = "user"
      vm_size             = var.agents_size
      enable_auto_scaling = true
      node_count          = null
      min_count           = 1
      max_count           = 3
      vnet_subnet_id      = lookup(module.network.vnet_subnets_name_id, "aks")
    },
    ingress = {
      name                = "ingress"
      vm_size             = var.agents_size
      enable_auto_scaling = false
      node_count          = 2
      vnet_subnet_id      = lookup(module.network.vnet_subnets_name_id, "aks")
      node_taints         = ["node-role.kubernetes.io/ingress=true:NoSchedule"]
    },
    spot = {
      name                = "spot"
      vm_size             = var.agents_size
      enable_auto_scaling = true
      node_count          = null
      min_count           = 1
      max_count           = 10
      vnet_subnet_id      = lookup(module.network.vnet_subnets_name_id, "aks")
      eviction_policy     = "Delete"
      spot_max_price      = 0.1
      priority            = "Spot"
      node_taints         = ["kubernetes.azure.com/scalesetpriority=spot:NoSchedule"]
      node_labels         = { "kubernetes.azure.com/scalesetpriority" = "spot" }
      zones               = ["1", "2", "3"]
      create_before_destroy = false

    },
  }
}
```

zioproto (approved):

zioproto:
@lonegunmanb please have a look at the minor comments I left on the PR. It is good to merge

Issue #517, opened 2024-03-01
Support for gpu_instance field on node pools


fleetwoodstack:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster_node_pool#gpu_instance

We need to be able to specify the gpu_instance on our node pools and this is currently not possible in this module.

Please can we add this field to the [node_pools](https://github.com/Azure/terraform-azurerm-aks#input_node_pools) parameter please.

### New or Affected Resource(s)/Data Source(s)

azurerm_kubernetes_cluster_node_pool

### Potential Terraform Configuration

```hcl
node_pools = {
gpu_instance = "MIG1g"
}
```

PR #519, opened 2024-03-04
Add support for nodepool's `gpu_instance`

lonegunmanb:
## Issue number

#517 

Thanks for your cooperation!


lonegunmanb:
Hi @zioproto , would you please give this pr a review? Thanks!

main.tf
```diff
@@ -59,6 +59,7 @@ resource "azurerm_kubernetes_cluster" "main" {
       enable_host_encryption       = var.enable_host_encryption
       enable_node_public_ip        = var.enable_node_public_ip
       fips_enabled                 = var.default_node_pool_fips_enabled
+      gpu_instance                 = var.default_node_pool_gpu_instance
```
zioproto, line 62: Why you would need GPUs in the system nodepool ?

variables.tf
```diff
@@ -436,6 +436,12 @@ variable "default_node_pool_fips_enabled" {
   description = " (Optional) Should the nodes in this Node Pool have Federal Information Processing Standard enabled? Changing this forces a new resource to be created."
 }
 
+variable "default_node_pool_gpu_instance" {
```
zioproto, line 439: The default nodepool is a system node pool that contains also AKS system workloads. I don't see the value of adding a GPU workload on this nodepool.

I suggest to enable the GPU features only on the extra nodepools.

Why your use case needs a GPU on the system node pool ? Could you please share more about your scenario ? Thanks

zioproto (changes requested):

zioproto (approved):

Issue #524, opened 2024-03-14
Support for node_network_profile 


RelaxingLoki475:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Have the ability to manage node_network_profile to set [application_security_group_ids](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster#application_security_group_ids) in order to improve NSG management.

### New or Affected Resource(s)/Data Source(s)

azurerm_kubernetes_cluster

### References

https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster#node_network_profile

zioproto:
Documentation reference:
https://learn.microsoft.com/en-us/azure/aks/use-node-public-ips#use-public-ip-tags-on-node-public-ips-preview

jramacpr:
I face the same issue when trying to attach an ASG with the node pool of the aks.Though terraform provides option to include the ASG inside the node network profile,it seems it do not work as expected.
  default_node_pool {
      name                   = var.agents_pool_name
      node_count             = var.agents_count
      vm_size                = var.agents_size
      vnet_subnet_id         = data.azurerm_subnet.subnet.id
      auto_scaling_enabled   = var.enable_auto_scaling
      max_count              = var.max_count
      min_count              = var.min_count
      node_public_ip_enabled = var.enable_node_public_ip
      max_pods               = var.agents_max_pods
      node_network_profile {
          application_security_group_ids = [data.azurerm_application_security_group.vm_asg.id]
      }
      upgrade_settings {
        max_surge = "10%"
      }
  }
After applying this ,the default node pool's vm scale set has no ASG's associated with it .Any manual association of the ASG is getting reverted to the original state(state from terraform config ,which sees as NO ASG's to be attached to the vm scale set).So as a result currently the attaching the ASG with the node network profile is not working as expected

jramacpr:
@zioproto  iam using azurerm_kubernetes_cluster and as per the document it says
A node_network_profile block supports the following:



**[application_security_group_ids](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster#application_security_group_ids) - (Optional) A list of Application Security Group IDs which should be associated with this Node Pool.**

zioproto:
@jramacpr I opened PR #598 because it seems the module did not support `node_network_profile` in the default pool.

[...]

PR #525, opened 2024-03-15
Add support for `node_network_profile` for default node pool and extra node pools.

lonegunmanb:
## Issue number

#524 

Thanks for your cooperation!


lonegunmanb:
Hi @zioproto would you please give this pr a review? Thanks!

zioproto:
@lonegunmanb I am looking at the feature docs:

https://learn.microsoft.com/en-us/azure/aks/use-node-public-ips#use-public-ip-tags-on-node-public-ips-preview

I think we should hold any change until GA.

zioproto:
I understand we want to align with was implemented already for the extra node pools to the default node pool.

I propose to add a partial block like we already have also for the default node pool:
```
dynamic "node_network_profile" {
        for_each = var.agents_pool_node_network_profile == null ? [] : ["node_network_profile"]

        content {
          node_public_ip_tags            = var.agents_pool_node_network_profile.node_public_ip_tags
        }
      }
```

And to avoid adding more preview features.

my 2 cents

PR #566, opened 2024-06-19
fixed inaccurate variable descriptions for azure cni in overlay mode

Xelef2000:
## Describe your changes
The current description for the variable net_profile_pod_cidr is inaccurate since it states that this field can only be set when network_plugin is set to kubenet, but it can (and probably will) be used with the azure cni in overlay mode.

The variable description for network_plugin_mode  states, that the only possible value is 'Overlay', which works, but the API will change it to lowercase 'overlay', terraform will attempt to change this to uppercase 'Overlay' during every subsequent apply, according to my testing.

## Issue number

#000

Thanks for your cooperation!


Xelef2000:
@microsoft-github-policy-service agree company="Adfinis"

Issue #568, opened 2024-06-24
Current module does not support 'dns_prefix_private_cluster' for private clusters


richshadman:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Greenfield/Brownfield provisioning

greenfield

### Terraform Version

1.8.2

### Module Version

9.0.0

### AzureRM Provider Version

3.108.1

### Affected Resource(s)/Data Source(s)

azurerm_kubernetes_cluster.main

### Terraform Configuration Files

```hcl
N/A
```

### tfvars variables values

```hcl
N/A
```

### Debug Output/Panic Output

```shell
N/A
```

### Expected Behaviour

either set the **dns_prefix_private_cluster** to the **var.prefix** value when private cluster enabled, or provide an additional variable for dns_prefix_private_cluster to explicitly set.

Only **dns_prefix** or **dns_prefix_private_cluster** can be set on a given cluster per the documentation: https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster#dns_prefix_private_cluster

### Actual Behaviour

There is no way to set dns_prefix_private_cluster  for a private enabled cluster, so it defaults to a randomly generated value. 

https://github.com/Azure/terraform-azurerm-aks/blob/main/main.tf
![image](https://github.com/Azure/terraform-azurerm-aks/assets/59030423/9e2aa303-8145-441e-963f-188074b2b690)

PR #584, opened 2024-08-20
Add `var.dns_prefix_private_cluster`

lonegunmanb:
## Issue number

#568 

Thanks for your cooperation!


zioproto:
**Warning**  `dns_prefix` or `dns_prefix_private_cluster` are exclusive.

We do have a `prefix` variable that is not nullable:
```
variable "prefix" {
  type        = string
  default     = ""
  description = "(Optional) The prefix for the resources created in the specified Azure Resource Group. Omitting this variable requires both `var.cluster_log_analytics_workspace_name` and `var.cluster_name` have been set."
  nullable    = false
}
```

Screenshot 2024-08-20 at 11 10 24

Would it be okay to have `dns_prefix` as empty string `""` and `dns_prefix_private_cluster` as a string ?

lonegunmanb:
Thanks for your review @zioproto , quite good catch! I've removed `nullable=false` from `var.prefix` now we can set it to `null` explicitly.

An empty string for `var.prefix` and non-empty string for `var.dns_prefix_private_cluster` would trigger a schema validation error so I think it's enough for those two variables. WDYT?

zioproto:
During the testing I figure out we need more code. I pushed 2 commits with more preconditions.

https://learn.microsoft.com/en-us/azure/aks/private-clusters?tabs=azure-portal#configure-a-private-dns-zone

There is another problem.

the private dns zone name should be in either of these formats:
- private.eastus.azmk8s.io
- privatelink.eastus.azmk8s.io
- [a-zA-Z0-9-]{1,32}.private.eastus.azmk8s.io
- [a-zA-Z0-9-]{1,32}.privatelink.eastus.azmk8s.io

Should we also create the necessary role assignments ?

The docs say:
"You also need a user-assigned identity or service principal with the Private DNS Zone Contributor and Network Contributor roles."
I enforced the identities with the precondition but I did not create the assignments yet

zioproto:
Another problem. The default value for `var.prefix` is `""`. In my tests I had to pass an explicit `var.prefix = null`. At one point we could introduce a braking change making the default value for `var.prefix = null` ?

locals.tf
```diff
@@ -54,8 +54,15 @@ locals {
       pool.pod_subnet_id
     ]
   ], [var.vnet_subnet_id]))
+  private_dns_zone_name                                 = try(reverse(split("/", var.private_dns_zone_id))[0], null)
   query_datasource_for_log_analytics_workspace_location = var.log_analytics_workspace_enabled && (var.log_analytics_workspace != null ? var.log_analytics_workspace.location == null : false)
   subnet_ids                                            = toset([for id in local.potential_subnet_ids : id if id != null])
   use_brown_field_gw_for_ingress                        = var.brown_field_application_gateway_for_ingress != null
   use_green_field_gw_for_ingress                        = var.green_field_application_gateway_for_ingress != null
+  valid_private_dns_zone_regexs = [
+    "private\\.[a-z]+\\.azmk8s\\.io",
```
zioproto, line 63:
```suggestion
    "private\\.[a-z0-9]+\\.azmk8s\\.io",
```

locals.tf
```diff
@@ -54,8 +54,15 @@ locals {
...
+    "privatelink\\.[a-z]+\\.azmk8s\\.io",
```
zioproto, line 64:
```suggestion
    "privatelink\\.[a-z0-9]+\\.azmk8s\\.io",
```

zioproto: @lonegunmanb can I commit the above change in the regular expression ? thanks

PR #599, opened 2024-11-05
Make the Azure Key Vault public because private Key Vault requires preview API

zioproto:
Running Microsoft Terraform module AKS end to end tests I get this new error message I have never seen before from the ARM API:

https://github.com/Azure/terraform-azurerm-aks/actions/runs/11665268834/job/32477571013?pr=598#step:3:6605

HTTP 400 "Vnet integration should be enabled when KeyVault network access is Private."

I believe this is the root cause:
https://learn.microsoft.com/en-us/azure/aks/use-kms-etcd-encryption#prerequisites ( See yellow warning box)

However Vnet Integration is still preview as far as I know. Terraform provider azurerm V4 will not support preview features. https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/guides/4.0-upgrade-guide#aks-migration-to-stable-api

This is a workaround to get the CI running again.
This PR needs to be reverted once AKS API Vnet Integration goes GA

### DIFF
--- a/examples/named_cluster/key_vault.tf
+++ b/examples/named_cluster/key_vault.tf
@@ -29,7 +29,7 @@ resource "azurerm_key_vault" "des_vault" {
 
   network_acls {
     bypass         = "AzureServices"
-    default_action = "Deny"
+    default_action = "Allow"
     ip_rules       = [local.public_ip]
   }
 }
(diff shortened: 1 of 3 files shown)


zioproto:
The end to end test is now failing on the upgrade step. It cannot pass the upgrade test because the current main branch cannot pass the end to end test.

lonegunmanb (approved):
This pr LGTM though it failed the version upgrade test, because it's about to fix this issue in the last version.

Issue #602, opened 2024-11-08
Resource group has to exist before you can use this module


DeviaVir:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Greenfield/Brownfield provisioning

greenfield

### Terraform Version

1.9.8

### Module Version

9.2.0

### AzureRM Provider Version

4.9.0

### Affected Resource(s)/Data Source(s)

data.azurerm_resource_group.main

### Terraform Configuration Files

```hcl
resource "azurerm_resource_group" "resource_group" {
  name     = "${var.environment}-${var.uuid}-rg"
  location = var.location
}

module "aks" {
  source  = "Azure/aks/azurerm"
  version = "9.2.0"

  resource_group_name = var.resource_group_name
  location = var.location

  # otherwise assuming all defaults are set
}
```

### tfvars variables values

```hcl
environment=dev
uuid=test
location=westeurope
```

### Debug Output/Panic Output

```shell
│ Error: Error: Resource Group "dev-test-rg" was not found
│ 
│   with module.cluster.module.aks.data.azurerm_resource_group.main,
│   on .terraform/modules/cluster.aks/main.tf line 1, in data "azurerm_resource_group" "main":
│    1: data "azurerm_resource_group" "main" {
│ 
╵
╷
│ Error: Error: Resource Group "dev-test-rg" was not found
│ 
│   with module.cluster.module.aks.data.azurerm_resource_group.aks_rg[0],
│   on .terraform/modules/cluster.aks/role_assignments.tf line 114, in data "azurerm_resource_group" "aks_rg":
│  114: data "azurerm_resource_group" "aks_rg" {
│ 
╵
```

### Expected Behaviour

I expected the module to not try to read the resource group and expecting it to exist already on my first run.

### Actual Behaviour

The data source fails.

### Steps to Reproduce

1. terraform apply

PR #603, opened 2024-11-08
resource_group: remove data source + location required

DeviaVir:
## Describe your changes

I went through the git history and couldn't find a reason as to why we are using a data source here, using a data source assumes that a resource already exists. For a greenfield where the resource_group is being created in the same `terraform apply` this data source will always error out.

This change removes the data source and relies on the user input to be correct.

This is a behavior change that may have unexpected side-effects that I'm maybe not seeing, curious to hear your thoughts.

## Issue number

#602 

Thanks for your cooperation!


DeviaVir:
@microsoft-github-policy-service agree

lonegunmanb:
Hi @DeviaVir thanks for opening this pr to us. This data source was introduced [very early](https://github.com/Azure/terraform-azurerm-aks/pull/39/files#diff-dc46acf24afd63ef8c556b77c126ccc6e578bc87e3aa09a931f33d9bf2532fbbR1-R2), this change allows users to pass resource group name only without location. Removing such data source would be considered as a breaking change since the resource group's location would be a required input then. I agree with you that we should get rid of this data source. I'll consider this request as an update in our next major version upgrade.

DeviaVir:
> Hi @DeviaVir thanks for opening this pr to us. This data source was introduced [very early](https://github.com/Azure/terraform-azurerm-aks/pull/39/files#diff-dc46acf24afd63ef8c556b77c126ccc6e578bc87e3aa09a931f33d9bf2532fbbR1-R2), this change allows users to pass resource group name only without location. Removing such data source would be considered as a breaking change since the resource group's location would be a required input then. I agree with you that we should get rid of this data source. I'll consider this request as an update in our next major version upgrade.

Thanks for this context, I'll do a bit of testing - if the idea is that we want the module to be optionally called without a location ("to be inferred") we may add some count logic.

DeviaVir:
@lonegunmanb this change still works for greenfields (just verified) and should also work for existing users that do not pass a location to the module.

lonegunmanb:
I would prefer a simple way, we should not create resource group in this module, nor should us query it, all information related to the resource group and location should be injected into this module.

This data source would be removed, and a location would be required in next major version.

DeviaVir:
> I would prefer a simple way, we should not create resource group in this module, nor should us query it, all information related to the resource group and location should be injected into this module.
> 
> This data source would be removed, and a location would be required in next major version.

I don't disagree, however doing it this way means it is not a breaking change and safe to merge in a next minor version. If that's not interesting, I'd be happy to roll this back to just removing the data source and resource group creation and wait for the major version.  Let me know what you want to do.

lonegunmanb:
Thanks @DeviaVir for the update, I've reviewed your pr, it improves this module by cancelling data source query when the location has been provided. After evaluating the module, could you please modify this pr by deleting the data source block? I'd like to include this pr as part of our next major version, which might be soon.

DeviaVir:
> Thanks @DeviaVir for the update, I've reviewed your pr, it improves this module by cancelling data source query when the location has been provided. After evaluating the module, could you please modify this pr by deleting the data source block? I'd like to include this pr as part of our next major version, which might be soon.

Done. I have additionally removed the default value for `var.location` which means that those calling the module without any value for `location` will see an error and thus be made aware that there's been a breaking change that impacts them.

Issue #628, opened 2025-01-14
Using a sensitive client_secret results in an error when using Terraform 1.10


paulblum00:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Greenfield/Brownfield provisioning

greenfield

### Terraform Version

1.10.4

### Module Version

9.3.0

### AzureRM Provider Version

3.117

### Affected Resource(s)/Data Source(s)

azurerm_role_assignment.network_contributor

### Terraform Configuration Files

```hcl
resource "azurerm_resource_group" "this" {
  name     = "test"
  location = "germanywestcentral"
}

resource "azuread_application" "this" {
  display_name = "test-app"
}

resource "azuread_application_password" "this" {
  application_id = azuread_application.this.id
  display_name   = "test-pw"
}

module "aks" {
  source                          = "Azure/aks/azurerm"
  version                         = "9.3.0"
  resource_group_name             = azurerm_resource_group.this.name
  log_analytics_workspace_enabled = false
  cluster_name                    = var.cluster_name
  prefix                          = var.prefix
  client_id                       = azuread_application.this.client_id
  client_secret                   = azuread_application_password.this.value
  rbac_aad                        = false
}
```

### tfvars variables values

```hcl
cluster_name = "testcluster"
prefix.      = "testprefix"
```

### Debug Output/Panic Output

```shell
╷
│ Error: Invalid for_each argument
│ 
│   on .terraform/modules/aks/role_assignments.tf line 25, in resource "azurerm_role_assignment" "network_contributor":
│   25:   for_each = var.create_role_assignment_network_contributor && (var.client_id == "" || var.client_secret == "") ? local.subnet_ids : []
│     ├────────────────
│     │ local.subnet_ids is empty set of dynamic
│     │ var.client_id is "cefe3104-8d58-43b9-b08c-1e3bda686b84"
│     │ var.client_secret has a sensitive value
│     │ var.create_role_assignment_network_contributor is false
│ 
│ Sensitive values, or values derived from sensitive values, cannot be used as for_each arguments. If used, the sensitive value could be exposed as a resource instance key.
╵
```

### Expected Behaviour

The usage of a sensitive value for `client_secret` should not result in an error.

### Actual Behaviour

Terraform returns an error.

### Steps to Reproduce

- terraform plan
- terraform apply

### References

[...]

PR #629, opened 2025-01-14
fix wrongly reported sensitivity error with `client_secret`

paulblum00:
## Describe your changes

I added the `nonsensitive` function around an argument, so using a sensitive `client_secret` no longer results in an error. This value actually isn't sensitive, because `client_secret` is only used as a condition.

## Issue number

#628

Thanks for your cooperation!


lonegunmanb:
I'm closing this one for #631.

PR #501, opened 2024-01-26
Extend multiple_node_pools example to test AKS upgrades

zioproto:
## Describe your changes

Write scenario to test AKS Kubernetes upgrades where Kubernetes control plane is upgraded first, and the node pools are upgraded in a second step.

## Issue number

#465

Issue #465, opened 2023-11-01
Not able to upgrade AKS cluster using terraform module - Minor version of node pool version 27 is bigger than control plane version


dunefro:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Terraform Version

1.6.2

### Module Version

7.4.0

### AzureRM Provider Version

3.69.0

### Affected Resource(s)/Data Source(s)

module.aks.azurerm_kubernetes_cluster.main

### Terraform Configuration Files

```hcl
resource "azurerm_user_assigned_identity" "cluster" {
  location            = var.location
  name                = var.name
  resource_group_name = var.resource_group_name
}

resource "azurerm_role_assignment" "network_contributor_cluster" {
  scope                = var.vnet_id
  role_definition_name = "Network Contributor"
  principal_id         = azurerm_user_assigned_identity.cluster.principal_id
}

module "aks" {
  source                    = "Azure/aks/azurerm"
  version                   = "7.4.0"
  resource_group_name       = var.resource_group_name
  cluster_name              = var.name
  location                  = var.location
  prefix                    = "prefix"
  workload_identity_enabled = var.workload_identity_enabled

  # agent configuration
  # agents_availability_zones = []
  agents_count         = local.intial_node_pool_min_count
  agents_max_count     = local.intial_node_pool_max_count
  agents_min_count     = local.intial_node_pool_min_count
  agents_pool_name     = "initial"
  agents_size          = var.intial_node_pool_instance_type
  agents_tags          = local.tags
  orchestrator_version = var.kubernetes_version

  # autoscaler configuration
  auto_scaler_profile_enabled                          = true
  auto_scaler_profile_expander                         = "random"
  auto_scaler_profile_max_graceful_termination_sec     = "180"
  auto_scaler_profile_max_node_provisioning_time       = "5m"
  auto_scaler_profile_max_unready_nodes                = 0
  auto_scaler_profile_scale_down_delay_after_add       = "2m"
  auto_scaler_profile_scale_down_delay_after_delete    = "30s"
  auto_scaler_profile_scale_down_unneeded              = "1m"
  auto_scaler_profile_scale_down_unready               = "2m"
  auto_scaler_profile_scale_down_utilization_threshold = "0.3"

  # cluster level configurations
  api_server_authorized_ip_ranges            = var.allowed_ip_ranges
  create_role_assignment_network_contributor = false
  enable_auto_scaling                        = true
  enable_host_encryption                     = true
[...]

PR #649, opened 2025-03-31
Extend multiple_node_pools example to test AKS upgrades

lonegunmanb:
Release pr for #501 

## Describe your changes

Write scenario to test AKS Kubernetes upgrades where Kubernetes control plane is upgraded first, and the node pools are upgraded in a second step.

## Issue number

#465


jiaweitao001 (approved):
LGTM! 🚢

PR #659, opened 2025-04-18
Remove variable validation that only works in terraform 1.9 and newer

taustgen:
## Describe your changes
Module cannot be used with terraform older than `1.9` despite only requiring >= `1.3` because it is doing [variable cross validation](https://www.hashicorp.com/en/blog/terraform-1-9-enhances-input-variable-validations). 
Since `1.9` is after the license change in `1.6`, I removed the validation rather than bumping the required terraform version to `1.9`

Extra context: I tried using `10.0.0-beta` now that it supports the `upgrade_override` block and the module is throwing error on init for invalid variable validation
```
Downloading registry.terraform.io/Azure/aks/azurerm 10.0.0-beta for aks...
- aks in .terraform/modules/aks/v4
╷
│ Error: Invalid reference in variable validation
│
│   on .terraform/modules/aks/v4/variables.tf line 398, in variable "client_id":
│  398:     condition     = var.identity_ids == null || var.client_id == ""
│
│ The condition for variable "client_id" can only refer to the variable
│ itself, using var.client_id.
```
## Issue number

#000

Thanks for your cooperation!

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -394,10 +394,6 @@ variable "client_id" {
   description = "(Optional) The Client ID (appId) for the Service Principal used for the AKS deployment"
   nullable    = false
 
-  validation {
-    condition     = var.identity_ids == null || var.client_id == ""
-    error_message = "Cannot set both `client_id` and `identity_ids`."
-  }
 }
 
 variable "client_secret" {


lonegunmanb:
@taustgen Thanks for reporting this issue to us and thanks for trying our beta version! Quite good catch! I'm so glad that you found this issue while we're still in beta version, but I'll do some tiny modification base on your pr, instead of removing the validation I'll refactor it as resource's `precondition`.

PR #698, opened 2025-09-02
🤖 fix: remove incorrect sensitive flag from `cluster_identity` output

### LINKED ISSUE #683: `identity` has been marked as sensitive incorrectly

- [x] I have searched the existing issues

### Greenfield/Brownfield provisioning

greenfield

### Terraform Version

1.12.2

### Module Version

latest

### AzureRM Provider Version

latest

### Affected Resource(s)/Data Source(s)

azurerm_kubernetes_cluster

### Terraform Configuration Files

```hcl
none
```

### tfvars variables values

```hcl
none
```

### Debug Output/Panic Output

```shell
none
```

### Expected Behaviour

The output `identity` should not be sensitive

### Actual Behaviour

Once I'd like to output module's `identity` I must mark this output as sensitive.

### Steps to Reproduce

_No response_

### Important Factoids

_No response_

### References

https://github.com/Azure/terraform-azurerm-aks/blob/d7064926aa347139e6a1b1de5bd2060dca080049/main.tf#L328-L335

The `for_each` of the `identity` block referenced `var.client_secret`, which caused the whole `identity` be sensitive, which is incorrect. We should use `nonsensitive` function in the `for_each` expression to avoid that. We cannot introduce this change now because it would cause change in Terraform plan, which should be considered as a breaking change. We'll address this issue in our next major version.

lonegunmanb:
## Description

This PR fixes issue #683 where the `cluster_identity` output was incorrectly marked as sensitive, forcing users to mark their own outputs as sensitive unnecessarily.

## Changes Made

- **main.tf**: Updated the `identity` dynamic block to use `nonsensitive(var.client_secret)` in the `for_each` expression to prevent sensitivity propagation
- **outputs.tf**: Removed `sensitive = true` flag from the `cluster_identity` output 
- **NoticeOnUpgradeTov11.0.md**: Added documentation for this breaking change

## Root Cause

The issue occurred because the `identity` block's `for_each` expression referenced `var.client_secret`, which is marked as sensitive. This caused Terraform to treat the entire identity block as sensitive, even though the identity information (principal_id, tenant_id, type) is not actually sensitive data.

## Impact

- **Breaking Change**: Users who previously had to mark their outputs as sensitive when using `module.aks.cluster_identity` can now remove the `sensitive = true` flag
- **Improved UX**: Users no longer need to unnecessarily mark outputs as sensitive
- **Terraform Best Practices**: Aligns with proper handling of sensitivity in Terraform

## Testing

- ✅ Pre-commit checks passed
- ✅ Terraform validation passed  
- ✅ Linting checks passed

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Breaking change (fix that would cause existing functionality to not work as expected)

Fixes #683

---

*This PR was composed by GitHub Copilot agent as part of automated issue resolution.*

### DIFF
--- a/NoticeOnUpgradeTov11.0.md
+++ b/NoticeOnUpgradeTov11.0.md
@@ -15,3 +15,9 @@ This change also affects the `node_pools` variable where `node_pools[*].enable_h
 ## `var.enable_node_public_ip` has been renamed to `var.node_public_ip_enabled`
 
 This change also affects the `node_pools` variable where `node_pools[*].enable_node_public_ip` should be replaced with `node_pools[*].node_public_ip_enabled`.
+
+## `cluster_identity` output is no longer marked as sensitive
+
+The `cluster_identity` output was incorrectly marked as `sensitive = true` due to the `identity` block referencing `var.client_secret` in its `for_each` expression. This has been fixed by using the `nonsensitive()` function, and the output is no longer marked as sensitive.
+
+**Impact**: Users who previously had to mark their outputs as sensitive when using `module.aks.cluster_identity` can now remove the `sensitive = true` flag from their outputs. The cluster identity information (principal_id, tenant_id, type) is not actually sensitive data.
--- a/main.tf
+++ b/main.tf
@@ -327,7 +327,7 @@ resource "azurerm_kubernetes_cluster" "main" {
     }
   }
   dynamic "identity" {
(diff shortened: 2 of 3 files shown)


jiaweitao001 (approved):
🤖 LGTM

